PR #17, opened 2015-04-20
NO-TICK - add backup policies

jamiegs:
Adds backup IAM policies to new mongo role creation

### DIFF
--- a/tyr/policies/ec2.py
+++ b/tyr/policies/ec2.py
@@ -18,3 +18,39 @@ allow_volume_control = """{
         }
      ]
 }"""
+
+allow_mongo_backup_snapshot = """{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Action": [
+                "ec2:CreateTags",
+                "ec2:DescribeTags",
+                "ec2:CreateSnapshot",
+                "ec2:DescribeInstances"
+            ],
+            "Sid": "Stmt1372359394000",
+            "Resource": [
+                "*"
+            ],
+            "Effect": "Allow"
+        }
+    ]
+}"""
+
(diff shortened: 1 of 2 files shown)


tyr/servers/mongo/node.py
```diff
@@ -8,7 +8,8 @@ class MongoNode(Server):
     CHEF_RUNLIST = ['role[RoleMongo]']
     CHEF_MONGODB_TYPE = 'generic'
 
-    IAM_ROLE_POLICIES = ['allow-volume-control']
+    IAM_ROLE_POLICIES = ['allow-volume-control', 'allow-mongo-backup-snapshot',
```
citruspi: Maybe reformat this like

```python
IAM_ROLE_POLICIES = [
    'allow-volume-control',
    'allow-mongo-backup-snapshot',
    'allow-mongo-snapshot-cleanup'
]
```

for better readability.

citruspi:
Aside from the formatting of `IAM_ROLE_POLICIES`, this looks good to me.

tyr/policies/ec2.py
```diff
@@ -18,3 +18,39 @@
         }
      ]
 }"""
+
+allow_mongo_backup_snapshot = """{
```
malkolm: This might be nitpicking but these are more general policies than just for mongo.  maybe a more broad name would be appropriate in case someone wanted to use the same privileges in another role.

jamiegs: Currently Mongo is the only thing that we backup using snapshots.  I could see if we do switch to doing other backups using snapshots that they would maybe require more or less privileges, such as not requiring tags.  I kind of like having it be specific to what Mongo backups need to run

Issue #13, opened 2015-03-19
Duplicate logging


wliao008:
Running this multiple time will produce duplicate logging.

>>> MongoDataNode(group='leroy', environment='stage', availability_zone='d', data_volume_size=150, data_volume_iops=0,mongodb_version='2.4.13').autorun()

16:11:29 [MongoDataNode] WARNING: No Instance Type provided
16:11:29 [MongoDataNode] WARNING: No Instance Type provided
16:11:29 [MongoDataNode] WARNING: No Instance Type provided
16:11:30 [MongoDataNode] INFO: Using Instance Type "m3.medium"
16:11:30 [MongoDataNode] INFO: Using Instance Type "m3.medium"
16:11:30 [MongoDataNode] INFO: Using Instance Type "m3.medium"
16:11:30 [MongoDataNode] INFO: Using group "leroy"
16:11:30 [MongoDataNode] INFO: Using group "leroy"
16:11:30 [MongoDataNode] INFO: Using group "leroy"
16:11:30 [MongoDataNode] INFO: Using type "mongo"
16:11:30 [MongoDataNode] INFO: Using type "mongo"
16:11:30 [MongoDataNode] INFO: Using type "mongo"
16:11:30 [MongoDataNode] INFO: Using Environment "stage"
16:11:30 [MongoDataNode] INFO: Using Environment "stage"
16:11:30 [MongoDataNode] INFO: Using Environment "stage"

PR #20, opened 2015-04-21
Prevent Duplicate Logging

citruspi:
See #13

This __only__ fixes duplicate logging, it does not add the new features discussed in the issue.


jamiegs:
looks good to me.

PR #23, opened 2015-04-27
Couchbase Automation

citruspi:
This adds Couchbase integration to `Tyr`. The following values are used

| Variable | Test/Stage Environment | Prod Environment |
|:----------:|:--------------------------------:|:-------------------------:|
| Chef Runlist | `role[RoleCache]` | `role[RoleCache], role[RoleSumoLogic]` |
| Instance Type | `m3.large` | `r3.large` |
| Bucket Name | `hudl-stage` | `hudl` |

The bucket name can be set to another value by passing in the argument `bucket_name`.

The `autorun()` function will actually create the bucket, allocating 100% of the cluster resources to it, after `chef-client` has finished running. This behavior differs from other server types, where `autorun()` finishes after setting the Chef attributes, and doesn't monitor `chef-client` or perform operations on the server.

### DIFF
--- a/tyr/servers/cache/server.py
+++ b/tyr/servers/cache/server.py
@@ -19,4 +19,21 @@ class CacheServer(Server):
                                             security_groups, block_devices,
                                             chef_path)
 
+    def configure(self):
+
+        super(CacheServer, self).configure()
+
+        # This is just a temporary fix to override the default security
+        # groups for Cache servers until the security_groups argument is
+        # removed.
+
+        self.security_groups = [
+            'management',
+            'chef-nodes',
+            self.envcl,
+            '{env}-cache-management'.format(env=self.environment[0])
+        ]
+
+        self.resolve_security_groups()
+
 
--- a/tyr/servers/cache/server.py
+++ b/tyr/servers/cache/server.py
@@ -9,9 +9,14 @@ class CacheServer(Server):
     def __init__(self, group=None, server_type=None, instance_type=None,
                     environment=None, ami=None, region=None, role=None,
                     keypair=None, availability_zone=None, security_groups=None,
-                    block_devices=None, chef_path=None):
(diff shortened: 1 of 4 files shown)


tyr/servers/cache/server.py
```diff
@@ -0,0 +1,152 @@
...
+            payload = {
+                'authType': 'sasl',
+                'bucketType': 'memcached',
+                'name': self.bucket_name,
+                'ramQuotaMB': memory_quota
+            }
+
+            auth = (username, password)
+
+            while True:
```
wliao008: Potential infinite loop? In our use case it's probably ok since most likely we'll be monitoring the execution.

citruspi: Yeah, it would loop until it managed to successfully connect to the Couchbase server. It's something we already do, specifically when managing the SSH connection and for automating stuff for MongoDB instances. I think it's fine since we're monitoring it and it's logging what's happening so it doesn't look like it's just stuck. The user would be aware of what's happening.

tyr/servers/server.py
```diff
@@ -701,6 +701,9 @@ def bake(self):
 
             self.chef_node.run_list = self.CHEF_RUNLIST
 
+            if self.chef_node.chef_environment == 'prod':
+                self.chef_node.run_list.append('role[RoleSumoLogic]')
```
malkolm: I think you won't need to do this.  In role_base::default, role_base::sumologic is included for all production nodes.  For Redhat nodes in EC2 this will add /var/log/messages and /var/log/secure to be indexed.  Per role_sumologic's readme, you can specify additional sources via data bag, but you won't need to include this to get the base stuff for a production node.

citruspi: Awesome, I wasn't aware of that. I've removed the inclusion.

PR #25, opened 2015-04-29
INF-730 - add vpc support

jamiegs:
Adds VPC support to Tyr.

In order to spin up a server in a VPC you must provide a subnet_id and exclude the availability zone.

When a subnet_id is supplied, tyr will now:
 - lookup the availability zone that the subnet is in
 - append a vpc suffix to the name of the security groups.  by default it will use -vpc.. you can override it for a VPC by setting the ```vpcSuffix``` tag on the VPC. for example: setting ```testvpc``` will append ```-testvpc``` to the end of the security groups.
      - this is to stop us from having security group name collisions.

- create a network adapter for the new server and apply security groups to the network adapter along with an ec2 generated public dns name

Also, You can't use security group names in VPCs, so this switches to resolve the security group id for both VPC and EC2-Classic. 

I also added *.pyc files to the .gitInore and fixed a few PEP 8 formatting issues and fixed a typo with setup.py where it was using ```include_package_date```  instead of ```include_package_data```

### DIFF
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -49,8 +49,6 @@ class MongoNode(Server):
         # groups for MongoDB nodes until the security_groups argument
         # is removed.
 
-
-
         self.resolve_security_groups()
 
     def run_mongo(self, command):
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -37,12 +37,18 @@ class MongoNode(Server):
 
         self.log.info('Using version {version} of MongoDB'.format(
                                                 version = self.mongodb_version))
+
+        # This is just a temporary fix to override the default security
+        # groups for MongoDB nodes until the security_groups argument
+        # is removed.
+
         self.security_groups = [
             'management',
             'chef-nodes',
             self.envcl,
             '{env}-mongo-management'.format(env=self.environment[0])
         ]
+
         self.resolve_security_groups()
 
     def run_mongo(self, command):
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -29,14 +29,14 @@ class MongoNode(Server):
 
     def configure(self):
 
-        super(MongoNode, self).configure()
-
         self.security_groups = [
             'management',
             'chef-nodes',
             self.envcl,
             '{env}-mongo-management'.format(env=self.environment[0])
         ]
+
+        super(MongoNode, self).configure()
 
         if self.mongodb_version is None:
             self.log.warn('MongoDB version not set')
@@ -45,11 +45,6 @@ class MongoNode(Server):
         self.log.info('Using version {version} of MongoDB'.format(
                                                 version = self.mongodb_version))
 
-        # This is just a temporary fix to override the default security
-        # groups for MongoDB nodes until the security_groups argument
-        # is removed.
-
-        self.resolve_security_groups()
 
     def run_mongo(self, command):
 
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -29,13 +29,6 @@ class MongoNode(Server):
 
     def configure(self):
 
-        self.security_groups = [
-            'management',
-            'chef-nodes',
-            self.envcl,
-            '{env}-mongo-management'.format(env=self.environment[0])
-        ]
-
         super(MongoNode, self).configure()
 
         if self.mongodb_version is None:
@@ -44,7 +37,13 @@ class MongoNode(Server):
 
         self.log.info('Using version {version} of MongoDB'.format(
                                                 version = self.mongodb_version))
-
+        self.security_groups = [
+            'management',
+            'chef-nodes',
+            self.envcl,
+            '{env}-mongo-management'.format(env=self.environment[0])
+        ]
+        self.resolve_security_groups()
 
     def run_mongo(self, command):
 
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -29,14 +29,14 @@ class MongoNode(Server):
 
     def configure(self):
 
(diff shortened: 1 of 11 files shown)


tyr/servers/mongo/node.py
```diff
@@ -24,10 +25,17 @@ def __init__(self, group = None, server_type = None, instance_type = None,
...
                                         security_groups, block_devices,
-                                        chef_path)
+                                        chef_path, subnet_id)
 
     def configure(self):
 
+        self.security_groups = [
+            'management',
+            'chef-nodes',
+            '{envcl}'.format(envcl=self.envcl),
```
citruspi: If the security group is just the value of `self.envcl`, you could just use `self.envcl` instead of formatting a string with that value.

tyr/servers/server.py
```diff
@@ -371,21 +386,58 @@ def blockdevicemapping(self):
...
 
+    def get_subnet_vpc_id(self, subnet_id):
+        vpcConn = VPCConnection()
+        subnets = vpcConn.get_all_subnets(
+            filters={'subnet_id': subnet_id})
+        if len(subnets) == 1:
+            vpc_id = subnets[0].vpc_id
+            return vpc_id
+        else:
+            self.log.error("None or greater then 1 subnet returned")
```
citruspi: If something happens which would cause the programme to exit, we should be using the `critical` log level.

tyr/servers/server.py
```diff
@@ -371,21 +386,58 @@ def blockdevicemapping(self):
...
+            exit(1)  # should probably throw an exception, if I knew how
```
citruspi: To raise a generic exception it would just be

```python
raise Exception("Something went wrong")
```

Alternatively you could add a new exception type to `tyr/servers/exceptions.py` and raise that.

``` python
# tyr/servers/exceptions.py

class NoSubnetReturned(Exception):
    pass

# tyr/servers/server.py

raise NoSubnetReturned("No subnet was returned")
```

tyr/servers/server.py
citruspi: I think `then` is supposed to be `than`.

citruspi: Based on the function name and code, I think you're passing in a subnet ID and returning a VPC ID. So perhaps the message would be `"None or greater than 1 VPC returned"`?

tyr/servers/server.py
```diff
@@ -371,21 +386,58 @@ def blockdevicemapping(self):
 
         return bdm
 
+    def get_subnet_vpc_id(self, subnet_id):
+        vpcConn = VPCConnection()
```
citruspi: Python recommends using

> lowercase with words separated by underscores as necessary to improve readability

for variable names as opposed to mixed case.

([PEP 008 - Method Names and Instance Variables][0])

[0]: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

tyr/servers/server.py
```diff
@@ -514,6 +566,23 @@ def establish_ec2_connection(self):
...
+        self.log.info(
+            "getting zone for subnet {subnet_id}".format(subnet_id=subnet_id))
+        VpcConn = VPCConnection()
+        filters = {'subnet-id': subnet_id}
+        subnets = VpcConn.get_all_subnets(filters=filters)
+
+        if len(subnets) == 1:
+            availability_zone = subnets[0].availability_zone
+
+            log_message = 'subnet {subnet_id} is in ' \
```
citruspi: Log messages should attempt to be consistent - capitalize the first word.

tyr/servers/mongo/node.py
```diff
@@ -41,13 +49,6 @@ def configure(self):
...
         # is removed.
 
-        self.security_groups = [
-            'management',
-            'chef-nodes',
-            self.envcl,
-            '{env}-mongo-management'.format(env = self.environment[0])
-        ]
-
         self.resolve_security_groups()
```
citruspi: If you set the security groups before calling `configure()` then they will automatically be resolved - no need to call it again.

tyr/servers/mongo/node.py
```diff
@@ -24,10 +25,17 @@ def __init__(self, group = None, server_type = None, instance_type = None,
...
-                                        chef_path)
+                                        chef_path, subnet_id)
 
     def configure(self):
 
+        self.security_groups = [
+            'management',
+            'chef-nodes',
+            '{envcl}'.format(envcl=self.envcl),
+            '{env}-mongo-management'.format(env=self.environment[0])
```
citruspi: Variables and properties like `self.envcl` and `self.environment` are being accessed at a time when they _may not_ be defined. If the user defines the `environment` argument there won't be an issue, but if they don't they'll get a `TypeError` because the `environment` variable isn't defined yet.

tyr/servers/server.py
```diff
@@ -532,17 +601,61 @@ def establish_route53_connection(self):
             self.log.error(str(e))
             raise e
 
+    def get_security_group_ids(self, security_groups):
+            ec2Conn = self.ec2
```
citruspi: Why make another copy of the connection instead of just using `self.ec2`?

tyr/servers/server.py
```diff
@@ -532,17 +601,61 @@ def establish_route53_connection(self):
...
+            security_group_ids = list()
```
citruspi: Another way to do this would be `security_group_ids = []`.

tyr/servers/server.py
citruspi: For the `get_security_group_ids` method, there are multiple possible log statements - one for the case where there are no matching security groups and one for the case where there are multiple matching security groups. Perhaps it would be good to do that here also? That way when an error occurs, the user knows what the problem was, not that it was one of two possible problems.

citruspi:
Looks good to me. :+1:

PR #26, opened 2015-04-29
Fixed the Fulla Volume Ownership

citruspi:
Instead of running `chown -R ec2-user:ec2-user /fulla/` in the `role_mongo` cookbook, we can set the desired `user` and `group` values to be used with the `hudl-ebs` cookbook when the volume is mounted.

This fixes the permissions issue that the `fulla` script was running into.


malkolm:
looks good to me.

Issue #24, opened 2015-04-28
Include Region Information in Name


citruspi:
Do we want to use `us-east-1`? Where would the region be in the name? We could put it before the availability zone, so hostnames would end up being `p-leroy-mongo-rs1-us-east-1-c-01.app.hudl.com`.

PR #29, opened 2015-04-30
Include Region Information in Instance Names

citruspi:
This includes region information in instances names along with availability zones.

Example names:

- `t-monolith-mongo-rs1-use1c-01`
- `t-monolith-cache-use1c-01`

### DIFF
--- a/tyr/servers/mongo/arbiter.py
+++ b/tyr/servers/mongo/arbiter.py
@@ -2,8 +2,8 @@ from member import MongoReplicaSetMember
 
 class MongoArbiterNode(MongoReplicaSetMember):
 
-    NAME_TEMPLATE = '{envcl}-rs{replica_set}-{zone}-arb'
-    NAME_SEARCH_PREFIX = '{envcl}-rs{replica_set}-{zone}-'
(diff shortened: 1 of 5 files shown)


citruspi:
This resolves #24.

tyr/servers/mongo/arbiter.py
```diff
@@ -2,8 +2,8 @@
 
 class MongoArbiterNode(MongoReplicaSetMember):
 
-    NAME_TEMPLATE = '{envcl}-rs{replica_set}-{zone}-arb'
-    NAME_SEARCH_PREFIX = '{envcl}-rs{replica_set}-{zone}-'
+    NAME_TEMPLATE = '{envcl}-rs{replica_set}-{location}-arb'
```
jamiegs: AWS seems refer to the region + availability zone as placement. do you think that'd be a better term?

jamiegs:
:+1: looks good to me

tyr/servers/mongo/arbiter.py
citruspi: The location isn't equivalent to the placement. The placement would be `us-east-1c` while the location would be `use1c`.

PR #30, opened 2015-05-01
Struct nginx server provisioning

malkolm:


tyr/servers/nginx/server.py
```diff
@@ -0,0 +1,197 @@
...
+        validation_key_file = open(validation_key_path, 'r')
+        validation_key = validation_key_file.read()
+
+        return template.format(hostname = self.hostname,
+                                fqdn = self.hostname,
+                                validation_key = validation_key,
+                                chef_server = 'http://chef.app.hudl.com/',
+                                logfile = '/var/log/chef-client.log')
+
+    def autorun(self):
```
citruspi: Since you're not modifying `autorun()` you shouldn't need to redefine it. `NginxServer` inherits from `Server` so this definition of `autorun` already exists.

tyr/servers/nginx/server.py
```diff
@@ -0,0 +1,197 @@
...
+            self.IAM_ROLE_POLICIES.append('allow-update-route53-stage')
+            self.IAM_ROLE_POLICIES.append('allow-modify-nginx-elbs-stage')
+            self.security_groups.append('s-nginx')
+        elif self.environment == 'prod':
+            self.IAM_ROLE_POLICIES.append('allow-update-route53-prod')
+            self.IAM_ROLE_POLICIES.append('allow-modify-nginx-elbs-prod')
+            self.security_groups.append('p-nginx')
+        self.resolve_iam_role()
+
+    def bake(self):
```
citruspi: Why redefine `bake()`?  I can see that the node's environment and run list is being set in the user data now, so it would probably just be defined twice if `bake()` wasn't redefined. First when `bake()` is called and then when the user data gets run. Which is fine, I don't think there's an issue with that. But if the `PyChef` API changed and we had to change the way that nodes were removed or added, then we would need to update both `Server` and `NginxServer`.

tyr/servers/nginx/server.py
malkolm: I thought i was modifying it since i'm not running self.route() and i'm not using wait=true on self.launch()

Is that not what I'm doing here?

citruspi: Whoops, sorry, I noted the absence of `wait=true` in the call to `launch()` but completely missed out on the fact that `route()` isn't called.

tyr/files/user_data.py
```diff
@@ -0,0 +1,100 @@
...
+
+cat << EOF >> /root/route53/upsert.sh
+/usr/bin/aws route53 change-resource-record-sets --region $MY_REGION --hosted-zone-id $ROUTE53_ZONE --change-batch file:///root/route53/upsert.json >> /root/route53/upsert.log 2>&1
+EOF
+
+chmod +x /root/route53/upsert.sh
+/root/route53/upsert.sh
+# /root/route53/upsert.sh can be run at any time to generate a new .json file
+# and upsert that record into Route 53.  This may be something to run every
+# time a node is started.  Useful if a node is stopped and started and the
```
jamiegs: I feel like we should be striving to making it easier to replace servers rather then making it easier to restart them.

citruspi:
Looks good to me. :+1:

PR #31, opened 2015-05-04
Added ReplicaSet Tag for MongoDB Replica Members

### LINKED ISSUE #22: EC2 Instance Tag With Replica Set Information
@jamiegs mentioned that we should have an EC2 instance tag on MongoDB nodes which would contain something like `MONOLITH-RS1`. If I remember correctly, this would be used when backing up MongoDB databases.

I'm not sure what the tag name would be. Probably `ReplicaSet`?

citruspi:
This resolves #22, adding an EC2 tag 'ReplicaSet' to MongoDB nodes which contains the name of the replica set.


jamiegs:
:+1: looks good to me

PR #32, opened 2015-05-05
EC2 Security Group Rule Automation

citruspi:
This adds support for automating the __addition__ of inbound rules to EC2 security groups. The modifications are __append only__, rules are not deleted.

It supports the re-use of a security group definition across multiple environments and groups through the use of regular expression rules and string formatting.

Sample definitions for Cache servers, MongoDB servers, and the Management and Chef-Nodes groups are included. These definitions have been used to successfully recreate the test environment security group for Cache servers from scratch and to spin up a MongoDB replica set in the stage environment for a new group without any work on the user's part.

The rule source can be expressed as the name of a security group, an IPv4 address, or a CIDR IP. Security Group ID's are not supported.


tyr/security_groups/chef_nodes.py
```diff
@@ -0,0 +1,11 @@
+rule = 'chef-nodes'
+
+definition = {
+    'rules': [
+        {
+            'port': 22,
+            'source': 'p-ops-vpn'
```
jamiegs: Why have this in chef-nodes also?  It's in management already

citruspi: You're right, I was just looking at the rules in `chef-nodes` and I didn't look for duplicate rules. I'll remove that.

tyr/security_groups/management.py
```diff
@@ -0,0 +1,33 @@
+rule = 'management'
+
+definition = {
+    'rules': [
+        {
+            'port': 22,
+            'source': [
+                'p-ops-vpn',
+                '216.229.9.101/32',
+                '76.84.137.182/32',
```
jamiegs: could we add all offices to all of these rules?

citruspi:
I've added support for `constants`, which are basically `variables`, so I'm not sure why I named them `constants`.

Anyway, they allow you to define values for strings which are then substituted in. `constants` always start with `@`. So, instead of having a rule

```python
{
    'port': 22,
    'source': [
        'p-ops-vpn',
        '216.229.9.101/32',
    ]
}
```

where you're not sure what `216.229.9.101/32` is for, you can now have a rule like

```python
{
    'port': 22,
    'source': [
        'p-ops-vpn',
        '@hudl-lincoln-east',
    ]
}
```

The value `216.229.9.101/32` will then be inserted when the rule is being compiled.

For what it's worth, I'm not sure why you'd want to do this, but the values can be anything - you could map a value to another security group.

tyr/security_groups/cache.py
```diff
@@ -0,0 +1,89 @@
...
+                    'rule': '[ts]-.+-cache',
+                    'value': '@hudl-lincoln-east'
+                },
+                {
+                    'rule': '[ts]-.+-cache',
+                    'value': '@hudl-omaha'
+                },
+                {
+                    'rule': '[ts]-.+-cache',
+                    'value': '@hudl-boston'
```
jamiegs: It might be cool to have an option of an `values` array to supply multiple.. It'd clean up a few of them.

tyr/security_groups/cache.py
```diff
@@ -0,0 +1,89 @@
...
+                {
+                    'rule': 's-.+-cache',
+                    'value': 'thors'
+                },
+                {
+                    'rule': 's-.+-cache',
+                    'value': '{env}-queues-jobs'
+                },
+                {
+                    'rule': 'p-.+-cache',
```
jamiegs: real minor but in some places you use rule first, then value.. which makes it a bit tougher to scan thru. having them being consistent would make it easier to read

tyr/security_groups/cache.py
citruspi: Yep, I thought of that yesterday. I just wasn't sure how much time I should spend on this pull request, but it shouldn't take too long to get that in.

tyr/servers/server.py
```diff
@@ -474,19 +477,115 @@ def resolve_security_groups(self):
             
             if not exists(group):
                 self.log.info('Security Group {group} does not exist'.format(
-                                group=group))
-                if self.subnet_id is None:
-                    self.ec2.create_security_group(group, group)
-                else:
-                    vpc_conn = VPCConnection()
-                    vpc_conn.create_security_group(
```
jamiegs: thanks for fixing this

tyr/servers/server.py
```diff
@@ -474,19 +477,115 @@ def resolve_security_groups(self):
...
                                 group=group))
 
+        for group in self.security_groups:
+            for key in security_groups.keys():
+                if not re.match(key, group): continue
+
+                self.log.info('Setting inbound rules for {group}'.format(
+                                                            group = group))
+
+                filters = {'group-name': group}
```
jamiegs: this loop is getting pretty long. could some of it be broken into functions?

tyr/security_groups/management.py
```diff
@@ -0,0 +1,35 @@
...
+        {
+            'port': '27017 - 28019',
+            'source': [
+                'p-ops-vpn'
+            ]
+        },
+        {
+            'port': 1433,
+            'source': [
+                '@hudl-lincoln-east'
```
jamiegs: these constants are great

tyr/security_groups/management.py
```diff
@@ -0,0 +1,35 @@
...
+            'source': [
+                'p-ops-vpn',
+                '@hudl-lincoln-east',
+                '@hudl-lincoln-west',
+                '@hudl-boston'
+                '@hudl-omaha'
+            ]
+        },
+        {
+            'port': 3389,
```
jamiegs: one other though.. it might be nice to have constants for ports.. or at least comments on them stating what they're for.

citruspi: So, technically comments are supported. You can just put Python comments in the middle of dictionaries

```
>>> d = {
...     'hello': 'world',
...     # testing
...     'foo': 'bar'
... }
>>> 
>>> d['hello']
'world'
>>> d['foo']
'bar'
```

but they're not in place. But more than adding comments, I definitely like the idea of constants for ports.

tyr/security_groups/management.py
```diff
@@ -0,0 +1,35 @@
...
+            'source': [
+                'p-ops-vpn',
+                '@hudl-lincoln-east',
+            ]
+        },
+        {
+            'port': '27017 - 28019',
```
jamiegs: hrm, this should probably be '27017 - 27019' + '28017 - 28019' there's 998 ports in there open that don't need to be open

PR #43, opened 2015-05-19
MongoDB Migration Hot Fixes

citruspi:
These are various fixes which are _already_ in use on the `p-infra-comm` box which were necessary to

1. Have the MongoDB migration module work with a VPC
2. Have the project run on a Linux machine

This:

- Adds support for waiting on DNS propagation when spinning up a node
- Establishes SSH connections over private IP
- Passes in SSH private key paths when establishing SSH connections
- Uses the private DNS name when claiming a replaced nodes DNS record

### DIFF
--- a/tyr/servers/server.py
+++ b/tyr/servers/server.py
@@ -812,8 +812,12 @@ named {name}""".format(path = d['path'], name = d['name']))
 
         while True:
             try:
+                keys = ['~/.ssh/stage', '~/.ssh/prod']
+                keys = [os.path.expanduser(key) for key in keys]
+
                 connection.connect(self.instance.private_dns_name,
-                                    username = 'ec2-user')
+                                    username='ec2-user',
+                                    key_filename=keys)
                 break
             except Exception:
                 self.log.warn('Unable to establish SSH connection')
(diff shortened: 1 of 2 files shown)


citruspi:
Note that once this is merged in, using the project will require that users

1. Are on the VPN
2. Have a copy of the stage SSH key located at `~/.ssh/stage` and the prod SSH key at `~/.ssh/prod`

tyr/servers/server.py
```diff
@@ -806,8 +812,12 @@ def connection(self):
 
         while True:
             try:
-                connection.connect(self.instance.public_dns_name,
-                                    username = 'ec2-user')
+                keys = ['~/.ssh/stage', '~/.ssh/prod']
```
malkolm: Does this mean we need our stage and prod ssh keys named as such?  Will it work if i'm using ssh-agent to add manage multiple ssh keys?

citruspi: It may work with `ssh-agent` on OS X even if the `key_filename` argument is used, but I haven't tested it. When the keys are added to the agent on Amazon Linux, it doesn't see them and requires that I pass in the key locations. Even if it does still check the agent, it would throw an error when those files don't exist.

malkolm: It looks good to me.  Just wanted to see if I needed to change my setup at all.

Issue #47, opened 2015-06-11
MongoDB Config Nodes Use the Incorrect Port


citruspi:
The MongoDB config nodes use port `27018`. They should be using `27019`.

PR #48, opened 2015-06-17
Change MongoDB Port for Config Servers to 27019

citruspi:
This fixes issue #47 which notes that config servers are currently configured to use `27018` for their MongoDB port. They should be using `27019`.


jamiegs:
:+1: looks good to me.

PR #52, opened 2015-06-30
Adding config for ScriptsServer node type

chrisgilbert:
Some additions for a ScriptServer type and a couple of new IAM groups needed.


tyr/servers/scripts/server.py
```diff
@@ -0,0 +1,41 @@
...
+
+        super(ScriptsServer, self).__init__(group, server_type, instance_type,
+                                            environment, ami, region, role,
+                                            keypair, availability_zone,
+                                            security_groups, block_devices,
+                                            chef_path, subnet_id, dns_zones)
+
+    def configure(self):
+        if not self.instance_type:
+            if self.environment == 'prod':
```
citruspi: Why check the environment if the same instance type is used regardless of the environment?

tyr/servers/scripts/server.py
```diff
@@ -0,0 +1,41 @@
...
+                self.instance_type = 't2.micro'
+            else:
+                self.instance_type = 't2.micro'
+        super(ScriptsServer, self).configure()
+
+        self.resolve_iam_role()
```
citruspi: This shouldn't be needed - it should be called as part of `configure()`.

tyr/servers/scripts/server.py
chrisgilbert: Good point..

tyr/servers/scripts/server.py
chrisgilbert: Yep, I see that, I will remove it.

PR #54, opened 2015-07-07
NO-TICK - add support for environment names to policies

jamiegs:
For many things we'll need different policies for different environments.  

This adds support for using an environment name in the name of the policy.

I've updated nginx's policies to use the environment rather then a if statement.


citruspi:
While this is a good idea, I'd recommend separating this into two pull requests. The first adding support for templated policy names and the second adding support for templated policy names to the Nginx server specifically.

The reason is that while the addition of support for templated policy names in general is fine, the integration with Nginx isn't. When we process IAM role policies, [we lookup the policy in the `policies` dictionary without checking if the key exists or not](https://github.com/hudl/Tyr/blob/struct-EnvironmentPolicies/tyr/servers/server.py#L554). With the previous `if` statement, the list of IAM role policies was only affected when the environment was `stage` or `prod`. Now, templated policies will be used regardless of the environment. 

So what would happen if someone spun up an Nginx server in the `test` environment? If it would fail to spin up because it would attempt to retrieve the policy `allow-modify-nginx-elbs-test` which isn't defined.

citruspi:
So whether or not you split this into two pull requests is your choice, but the complete pull request as it stands right now wouldn't work in the `test` environment.

jamiegs:
I'll add the test environment policies. I don't feel like splitting it at this point would really gain us anything.

citruspi:
> I don't feel like splitting it at this point would really gain us anything.

That's fair. I only suggested that in case we wanted to merge the support for templated policies in sooner if other pull requests are depending on it. But the pull request is small enough that it shouldn't be an issue.

> I'll add the test environment policies.

So, that's one option. I had a couple other ideas. My first thought was that we could just fail silently if a policy doesn't exist, log an error, and carry on. I'm not comfortable with that because if a policy does need to be added and it isn't for whatever reason, I'd rather it fail to run than just carry on. My other thought would be to add support for policies whose value is `None`. So in the policies dictionary you'd have


```python
{
    ...
    'allow-modify-nginx-elbs-test': None,
    'allow-modify-nginx-elbs-stage': elb.allow_modify_nginx_elbs_stage,
    'allow-modify-nginx-elbs-prod': elb.allow_modify_nginx_elbs_prod
    ...
}
```

Then, when policies are being processed, check if the value is `None`. If so, skip it. But that raises the question of what to do if a policy is defined on a role, but the value in the dictionary is set to `None`. Should the policy defined on the role be deleted?

One other option would be to replace

```python
if self.environment == 'stage':
    self.IAM_ROLE_POLICIES.append('allow-modify-nginx-elbs-stage')
elif self.environment == 'prod':
    self.IAM_ROLE_POLICIES.append('allow-modify-nginx-elbs-prod')
```

with

```python
if self.environment == 'stage' or self.environment == 'prod':
    self.IAM_ROLE_POLICIES.append('allow-modify-nginx-elbs-{environment}')
```

Then you don't need to worry about policies that don't exist and you don't need to add policies simply to make sure every environment is accounted for. The templated policy would only be used in applicable environments.

You could even make it a little cleaner with

```python
if self.environment in ['stage', 'prod']:
    self.IAM_ROLE_POLICIES.append('allow-modify-nginx-elbs-{environment}')
```

citruspi:
Looks good to me. :+1:

tyr/servers/server.py
```diff
@@ -545,18 +545,26 @@ def resolve_iam_role(self):
         self.log.info('Existing policies: '
                       '{policies}'.format(policies=existing_policies))
 
-        for policy in self.IAM_ROLE_POLICIES:
+        for policy_template in self.IAM_ROLE_POLICIES:
+            policy = policy_template.format(environment = self.environment)
```
citruspi: There are extra spaces around the `=`.

PR #57, opened 2015-07-11
Add pep8 compliance

msambol:
I made some stylistic changes so that the project now passes pep8,
except for the use of lambda functions, which is too damn clean.


citruspi:
This is some awesome stuff, thanks for the contribution!

PR #60, opened 2015-07-21
Allow for Multi-AZ Mongo Cluster in a VPC

alexdebrie:
This pull requests allows for the creation of a Multi-AZ Mongo Cluster within a VPC. Previously, you could only spin up a Multi-AZ cluster in EC2-Classic.

Boto's VPC module doesn't have the most elegant API. For example, you can't use a VPC id to grab all subnets for that particular VPC. You need to grab all subnets for your account, then you can check the `vpc_id` on each subnet. 

I tried to limit the damage and stay close to Tyr's style, but I'm open to suggestions. I tested both a cluster within a VPC and without, and they both worked.

### DIFF
--- a/tyr/clusters/mongo.py
+++ b/tyr/clusters/mongo.py
@@ -2,6 +2,7 @@ import logging
 from tyr.servers.mongo import MongoDataNode, MongoArbiterNode
 import time
 import json
+import boto
 
 
 class MongoCluster(object):
@@ -22,7 +23,7 @@ class MongoCluster(object):
                  replica_set=None, security_groups=None,
                  block_devices=None, data_volume_size=None,
                  data_volume_iops=None, data_nodes=None,
-                 mongodb_version=None):
+                 mongodb_version=None, vpc_id=None):
 
         self.nodes = []
 
@@ -43,6 +44,7 @@ class MongoCluster(object):
         self.data_nodes = data_nodes
         self.mongodb_version = mongodb_version
         self.dns_zones = dns_zones
+        self.vpc_id = vpc_id
 
     def provision(self):
 
@@ -53,6 +55,26 @@ class MongoCluster(object):
         while len(zones) < (self.data_nodes+1):
 
             zones += zones
+
+        subnets = []
+
+        if self.vpc_id:
+            
+            self.log.info("VPC ID provided. Checking for subnets.")
+
+            conn = boto.connect_vpc()
+            all_subnets = conn.get_all_subnets()
+
+            subnets = [subnet.id for subnet in all_subnets
+                        if subnet.vpc_id == self.vpc_id]
+
+        if not subnets:
+            
+            subnets = [None]
+
+        while len(subnets) < (self.data_nodes+1):
+
+            subnets += subnets
 
         self.log.info('Provisioning MongoDB Data Nodes')
 
@@ -74,7 +96,8 @@ class MongoCluster(object):
                                  availability_zone=zones[i],
                                  data_volume_size=self.data_volume_size,
                                  data_volume_iops=self.data_volume_iops,
-                                 mongodb_version=self.mongodb_version)
+                                 mongodb_version=self.mongodb_version,
+                                 subnet_id=subnets[i])
 
             node.autorun()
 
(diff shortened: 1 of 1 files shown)


tyr/clusters/mongo.py
```diff
@@ -54,6 +56,26 @@ def provision(self):
...
+        subnets = []
+
+        if self.vpc_id:
+            
+            self.log.info("VPC ID provided. Checking for subnets.")
+
+            conn = boto.connect_vpc()
+            all_subnets = conn.get_all_subnets()
+
+            subnets = [subnet.id for subnet in all_subnets
```
citruspi: So while I can definitely appreciate the flexibility that dynamically populating the list of subnets brings us, I have a slight fear of this coming back to bite us. So a couple notes:

- we only use the availability zones `c`, `d`, and `e`
- a VPC can have multiple subnets using the same availability zone

So we would run into an issue if a new subnet was created. The new subnet might just be created for testing something or to isolate some instances in a separate subnet, but all of a sudden instances spun up with `MongoCluster` would automatically be placed inside the subnet.

If an experimental subnet were created for zone `a` or `b`, then what happened would be pretty apparent in the logs - you'd see an instance with a name like `p-monolith-mongo-use1a-01.app.hudl.com`. But what if the new subnet was created in zones `c`, `d`, or `e`? You'd see what looks like an expected instance name and might not realize what had happened if

- you don't spot the different subnet ID in the logs when the instance is being provisioned
- you don't spot the private IP address being placed in a new CIDR block

So basically, I'm worried about someone creating a new subnet which may not be intended for general use, `MongoCluster` automatically electing to place nodes there, and no one realizing what's happened.

It's not as flexible, but I think I might just lean towards hard coding it and then updating the list when we create a new subnet which is intended for use. Instead, or in addition, you could have an argument `subnet_ids` in addition to `vpc_id` so that developers can customize the subnets being used if they need to. If the value is `None`, use the hardcoded values, if not, use the list provided.

jamiegs: I agree, we'd need to be able to pass in a list of subnets

alexdebrie:
[Latest commit](https://github.com/hudl/Tyr/commit/195f04647a63d34f37ae7d695f8a2f5311e12b29) removes the `vpc_id` option and instead allows a user to pass in a comma-separated list of subnet IDs in the `subnet_ids` argument. Let me know if either of you would like any changes.

tyr/clusters/mongo.py
```diff
@@ -2,6 +2,7 @@
 from tyr.servers.mongo import MongoDataNode, MongoArbiterNode
 import time
 import json
+import boto
```
citruspi: Since we're no longer querying Amazon VPC for the subnets associated with a VPC, this import is no longer necessary.

tyr/clusters/mongo.py
```diff
@@ -43,6 +44,8 @@ def __init__(self, group=None, server_type=None, instance_type=None,
         self.data_nodes = data_nodes
         self.mongodb_version = mongodb_version
         self.dns_zones = dns_zones
+        if subnet_ids:
+            self.subnet_ids= [subnet_id.strip() for subnet_id in subnet_ids.split(',')]
```
citruspi: So it looks like you're expecting the value of `subnet_ids` to be a string like `"subnet-abc,subnet_def,subnet_ghi"`, give or take some spaces. I think we should just expect an actual list `["subnet-abc", "subnet_def", "subnet_ghi"]`.

alexdebrie:
Fixed to have `subnet_ids` be a list and removed boto. Worked on a quick test run.

tyr/clusters/mongo.py
```diff
@@ -54,6 +54,14 @@ def provision(self):
 
             zones += zones
 
+        if len(self.subnet_ids) < 1:
```
citruspi: Just a thought. The length of a list will never be negative, so the only possible case where

```python
len(self.subnet_ids) < 1
```

is `True` is when the list is empty. It might be better to write it as

```python
if len(self.subnet_ids) == 0
```

so you don't have to take the time to think about all the cases where it would be `True`.

citruspi:
Aside from my comment on 

```python
len(self.subnet_ids) < 1
```

this looks good to me. :+1:

PR #61, opened 2015-07-21
adding hint into bootstrap so we get ec2 attributes on 1st chef run

malkolm:
I was doing this in role_base, but if this hint file ec2.json doesn't already exist before the 1st chef run the attributes don't populate.  If it's in role_base it takes 2 runs to get the attributes.


citruspi:
Looks good to me. :+1:

PR #62, opened 2015-07-22
MongoDB Compaction Utility

citruspi:
This utility performs compaction on a MongoDB replica set. It will accept any member of a replica set, determine the rest of the members and proceed.

### Procedure

Here are the steps it follows:

1. Determine members of the replica set
2. Determine primary and secondary (and ignore arbiters).
3. Determine if every secondary is syncing to the primary. If not use `rs.syncFrom` to instruct the secondary to.
4. On every secondary
    1. Remove the file `/home/ec2-user/compact.js` if it exists.
    2. Download a fresh copy of the compaction script from Amazon S3.
    3. Place the secondary into Maintenance Mode in Stackdriver.
    4. Run the script.
    5. Wait until the secondary is no longer recovering before continuing.
    6. Remove the secondary from Maintenance Mode.
5. Determine the primary.
6. Instruct the primary to step down.
7. Wait until the election is finished.
8. Determine if every secondary is syncing to the new primary. If not use `rs.syncFrom` to instruct the secondary to.
9. On the previous master
    1. Remove the file `/home/ec2-user/compact.js` if it exists.
    2. Download a fresh copy of the compaction script from Amazon S3.
    3. Place the secondary into Maintenance Mode in Stackdriver.
    4. Run the script.
    5. Wait until it is no longer recovering before continuing.
    6. Remove it from Maintenance Mode.

Any collection that is not a "system" collection will be compacted - the compaction script will ignore the following databases:
- `admin`
- `local`
- `test`

and any collections in other databases that start with `system` and compact everything else.

### Usage

`tyr.utilities.compact_mongo_collection.compact_mongodb_server` accepts three arguments:

| Argument | Description | Default Value |
|:--------------|:---------------|:-------------------|
| host | Any member of the replica set which should be compacted. | |
| `version` | The version of the compacting script to be used. | |
| `prompt_before_failover` | Should the utility wait for the user to press `return` before failing over the replica set to compact the primary. | `True` |

There is also a `compact-mongodb-servers` command which accepts a JSON or YAML file containing a list of dictionaries with these arguments. For example

```yaml
# conf.yaml
---
- host: s-leroy-mongo-rs1-c-01.app.staghudl.com
  version: '2012-06-27'
  prompt_before_failover: false
```

would __compact every member of `s-leroy-mongo-rs1`__ using version __2012-06-27__ of the compaction script, __without__ waiting for the user to hit `return` before stepping down the primary.


jamiegs:
Just to verify, it syncs only 1 secondary at a time right?  And it does include fulla nodes too?

citruspi:
@jamiegs Yes and yes.

citruspi:
Well, it will compact - not sync - them, but yeah.

Z3r0Sum:
@citruspi I just had the question we spoke about in person:

It doesn't seem like there's any flow control around determining the primary.  I was curious about what happens if the primary was null (even though the 120 second sleep seems safe based on your prior experience)?

```python
def determine_primary(self, member):

        log.debug('Using db.isMaster() to determine the primary')
        response = run_mongo_command(member, 'db.isMaster()')
        primary = response['primary'].split(':')[0]

        self.primary = primary

        log.debug('The primary is {primary}'.format(primary=self.primary))
```

jamiegs:
Seems good to me. Did you have to make any changes to the compact.js script? Do we need to fork it or anything to store our changes somewhere?

jamiegs:
I accidentally clicked the wrong button and closed it instead of commenting

citruspi:
@jamiegs Nope, no changes made.

jamiegs:
@citruspi Nice! I hope that script made things a bit easier for you.

citruspi:
@Z3r0Sum so I'm not sure what the output of `db.isMaster()` looks like when there is no master. If it just throws an error that isn't JSON formatted, then

```python
primary = response['primary'].split(':')[0]
```

would throw a `TypeError`.

If it throws a JSON formatted error or if the `primary` key was simply omitted, that line would throw a `KeyError`. 

If it returns JSON which has the `primary` key defined, `self.primary` would be set to that value and when an attempt to interact with the replica set is made in the future, we would probably get errors about it being unable to SSH into the primary.

Either way, it's probably something we should handle, so thanks for bringing it up. I think it's out of scope for this pull request, but feel free to open a new issue.

@jamiegs Thoughts? Should we wait for that fix to take place and get merged in before merging this or will we be fine without it?

citruspi:
A fix for the case where `db.isMaster()` returns invalid information or throws an error if the primary is missing has been resolved in #64.

Issue #63, opened 2015-07-22
Account for being unable to determine the primary within mongodb


Z3r0Sum:
It doesn't seem like there's any flow control around determining the primary. I was curious about what happens if the primary was null (even though the 120 second sleep seems safe based on your prior experience)?

```python
def determine_primary(self, member):

        log.debug('Using db.isMaster() to determine the primary')
        response = run_mongo_command(member, 'db.isMaster()')
        primary = response['primary'].split(':')[0]

        self.primary = primary

        log.debug('The primary is {primary}'.format(primary=self.primary))
```

Insight from @citruspi:

```python 
primary = response['primary'].split(':')[0]
```
would throw a TypeError.

If it throws a JSON formatted error or if the primary key was simply omitted, that line would throw a KeyError.

If it returns JSON which has the primary key defined, self.primary would be set to that value and when an attempt to interact with the replica set is made in the future, we would probably get errors about it being unable to SSH into the primary.

PR #64, opened 2015-07-22
Handle Error Cases when Accessing the Primary using db.isMaster()

citruspi:
Previous discussion from #62 and #63.

> @Z3r0Sum so I'm not sure what the output of `db.isMaster()` looks like when there is no master. If it just throws an error that isn't JSON formatted, then

> ```python
primary = response['primary'].split(':')[0]
```

> would throw a `TypeError`.

> If it throws a JSON formatted error or if the `primary` key was simply omitted, that line would throw a `KeyError`. 

> If it returns JSON which has the `primary` key defined, `self.primary` would be set to that value and when an attempt to interact with the replica set is made in the future, we would probably get errors about it being unable to SSH into the primary.

This does not handle the case where the `primary` property is defined and is a string, but not actually a valid primary.

### DIFF
--- a/tyr/utilities/replace_mongo_server.py
+++ b/tyr/utilities/replace_mongo_server.py
@@ -56,15 +56,15 @@ class ReplicaSet(object):
         try:
             primary = response['primary'].split(':')[0]
         except TypeError as e:
-            log.critical(e)
+            log.critical(str(e))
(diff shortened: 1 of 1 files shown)


jamiegs:
looks good to me.

PR #67, opened 2015-07-27
NO-TICK - add eureka support

jamiegs:
Adds support for Eureka servers to tyr. 

These servers require the ability to describe instances and tags to get its state from EC2.


citruspi:
This looks good to me. :+1:

PR #70, opened 2015-07-30
Revert "Revert "Add IISNode IICluster and Autoscaler classes""

chrisgilbert:
A revert of a revert is confusing - anyhow, this is the original pull request reopened to add the IISNode and Autoscaler types.


tyr/clusters/autoscaling.py
```diff
@@ -0,0 +1,112 @@
...
+                 autoscaling_group,
+                 node_obj,
+                 desired_capacity=1,
+                 max_size=1,
+                 min_size=1,
+                 default_cooldown=300,
+                 availability_zones=None,
+                 subnet_ids=None,
+                 health_check_grace_period=300):
+        self.log = logging.getLogger('tyr.clusters.AutoScaler')
```
citruspi: Each component should be capitalized.

tyr/clusters/iis.py
```diff
@@ -0,0 +1,104 @@
...
+                 autoscaling_group=None,
+                 desired_capacity=1,
+                 max_size=1,
+                 min_size=1,
+                 default_cooldown=300,
+                 availability_zones=None,
+                 health_check_grace_period=300,
+                 launch_configuration=None):
+
+        self.log = logging.getLogger('tyr.clusters.IISCluster')
```
citruspi: Each component should be capitalized.

tyr/servers/server.py
```diff
@@ -54,19 +63,19 @@ def establish_logger(self):
         except:
             pass
 
-        log = logging.getLogger(self.__class__.__name__)
+        log = logging.getLogger('tyr.{c}'
```
citruspi: Each component should be capitalized.

tyr/servers/iis/node.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.ports_to_authorize = [9000, 9001, 8095, 8096]
+
+        self.IAM_ROLE_POLICIES.append('allow-web-initialization-{0}'
+                                      .format(self.environment))
+        self.IAM_ROLE_POLICIES.append('allow-outpost-sns-{0}'
+                                      .format(self.environment))
+
+    def configure(self):
+        super(IISNode, self).establish_logger()
```
citruspi: Why call `establish_logger` as part of `configure`? It should already be called as part of `autorun`.

tyr/clusters/iis.py
```diff
@@ -0,0 +1,104 @@
...
+        if self.availability_zones:
+            self.node_zone = availability_zones[0]
+        else:
+            self.node_zone = None
+
+    def provision(self):
+
+        self.log.info('Provisioning IISCluster')
+        if not self.launch_configuration:
+            self.launch_configuration = "{0}-{1}-web".format(
```
citruspi: Named placeholders are preferred to positional indexes.

tyr/clusters/iis.py
```diff
@@ -0,0 +1,104 @@
...
+                          default_cooldown=self.default_cooldown,
+                          availability_zones=self.availability_zones,
+                          subnet_ids=self.subnet_ids,
+                          health_check_grace_period=self.
+                          health_check_grace_period,
+                          node_obj=node)
+        auto.autorun()
+
+    def baked(self):
+        return all([node.baked() for node in self.nodes])
```
citruspi: 1. The nodes will never be baked.
2. If you call this, you'll be stuck in an infinite loop.
3. `self.nodes` isn't defined. So, you probably won't actually be stuck in an infinite loop, it'll just crash.

tyr/servers/iis/node.py
```diff
@@ -0,0 +1,100 @@
...
+        if self.mongos_service:
+            mongo_ops = ("MongosHost", "Disabled", "MongosService")
+            if self.mongos_service not in mongo_ops:
+                raise ValueError(
+                    "Mongo service name must be one of: {0}".format(
+                        mongo_ops))
+
+        self.ports_to_authorize = [9000, 9001, 8095, 8096]
+
+        self.IAM_ROLE_POLICIES.append('allow-web-initialization-{0}'
```
jamiegs: we've now got the ability to add {environment} into the name of the policy and it'll handle replacing it.  see https://github.com/hudl/Tyr/pull/54/files

tyr/servers/iis/node.py
chrisgilbert: Yes, I think this is because when creating a node object as a template for the autoscale group, you never actually call autorun.  It's mostly because of the way the node objects are created (you don't want to call autorun on them when building an autoscale cluster, because that launches the node itself) so instead, I call it in configure to to avoid losing any logging.  

In any case, establish_logger can be called multiple times safely, so this was a work around for the slightly odd way this node.py object behaves when added to the autoscale group.  If you create it as a standalone node, it should work the normal way, but it means it will also do logging correctly if you only instantiate a node object and pass to the IISCluster to create the autoscale group.  Does that make sense?

tyr/clusters/iis.py
chrisgilbert: You're right, they will never bake.  I should just return false instead I guess?

tyr/servers/iis/node.py
chrisgilbert: OK, I have changed that.  It won't work for the security groups though right?

chrisgilbert:
I have tested this again as so:
```
from tyr.clusters.iis import IISCluster
i = IISCluster(group='chrisgtest', environment='stage', ami='ami-b81801d0', subnet_ids=['subnet-9a41d9c3'], desired_capacity=1, min_size=0, max_size=3)
i.autorun()
```
This works fine, though the server doesn't register with eureka, I think because it's a made up group without any cluster config yet.  The only thing that seemed to be missing was the tag creation permission which I have fixed now.  But you might want to have a look yourselves and see if I missed anything.

chrisgilbert:
P.S, to create an autoscale group with several AZs, you just need to specify the extra subnets in the list.  I've been using one instance for testing here, but upping the desired numbers and adding the extra subnets will add them to the ASG as you would expect.

chrisgilbert:
Hi Mihir, I've fixed those positional placeholders and resolved the merge conflicts from master.  Let me know if you are OK with me merging this now.

chrisgilbert:
OK, am going to merge this now.   Please shout if there are any more problems to look at - I have addressed all the ones raised up to now, and tested things with a new merge from latest master.

PR #71, opened 2015-07-30
StackDriver Alerts - Allow Automatic Creation of stackdriver alerts for groups

chrisgilbert:
Here is a spreadsheet with current alerts applied to all groups.  There are some limitations to what this does at the moment, I haven't looked at alerts attached to instances (there aren't many anyway), but I guess that should be in here.  

https://docs.google.com/spreadsheets/d/1VR0uq5mzQRuzL-NhG0BXxMqev5M-zN6Ysb3kqbZXPtI/edit?usp=sharing

See the https://github.com/hudl/Tyr/blob/STRUCT-AddStackDriverAlerts/tyr/alerts/README.md for some examples on testing which alerts are applied and creating new ones.


tyr/alerts/stackdriver.py
```diff
@@ -0,0 +1,257 @@
...
+        self.read_config()
+        self.policies = self.get_policy_list()
+        self.groups = self.get_group_list()
+
+    def exit_flush(self, code):
+        logging.shutdown()
+        exit(code)
+
+    def read_config(self):
+        with open(os.path.abspath(os.path.dirname(__file__)) +
```
citruspi: I'd recommend using something like `tyr.policies` instead of attempting to read in a JSON file like this.

tyr/alerts/stackdriver.py
```diff
@@ -0,0 +1,257 @@
...
+                  '/../policies/stackdriver_conditions.json', 'r') as jsonfile:
+            self.conditions = json.load(jsonfile)
+
+        with open(os.path.abspath(os.path.dirname(__file__)) +
+                  '/../policies/stackdriver_notifications.json',
+                  'r') as jsonfile:
+            self.notifications = json.load(jsonfile)
+
+    def get_policy_list(self):
+        r = requests.get(API_ENDPOINT + LIST_POLICY_URL, headers=self.headers)
```
citruspi: You should probably check the status code to make sure the request was successful.

tyr/alerts/stackdriver.py
```diff
@@ -0,0 +1,257 @@
...
+        return r.json()
+
+    def pretty_print_json(self, obj):
+        print(json.dumps(obj, sort_keys=True,
+                         indent=4, separators=(',', ': ')))
+
+    def get_group_list(self):
+        groups = requests.get(API_ENDPOINT + LIST_GROUPS_URL,
```
citruspi: You should probably check the status code to make sure the request was successful.

tyr/alerts/stackdriver.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def get_group_id_by_name(self, name):
+        for group in self.groups:
+            if group['name'] == name:
+                log.debug('Retrieved group {id} for {group}'
+                          .format(id=group['id'], group=name))
+                return group['id']
+
+    def get_group_by_id(self, id):
+        group = requests.get(API_ENDPOINT + GET_GROUP_URL.format(group_id=id),
```
citruspi: You should probably check the status code to make sure the request was successful.

citruspi:
What exactly will this be used for? How and when will it be used?

chrisgilbert:
The idea is that we will be able to add alerts when adding new servers from Tyr, so we need a way of defining what those policies should be in config, and being able to add them to new servers.

Stackdriver's API doesn't allow creation of alerts yet though, so that requires a bit of hackery to send a JSON object to the right URL (see the unfinished method at the bottom for a rough idea).

The first problem though is just working out what alerts we have and what groups they are attached to.  It's difficult to see whether alerts are active for what groups, because their API gives a list of policies, with embedded group IDs for each one.  So most of what I have done so far is focussed on how to see what is in there already.  Some of these methods can come out in order to integrate into Tyr later.

You can see in the speadsheet what the current list looks like.  One of the problems is that we are using inconsistent metric types - sometimes cloudwatch, sometimes winagent, sometimes agent:aggregation etc.  So I think we just need to know what type of metrics we should apply to which groups first.

tyr/policies/stackdriver_conditions.json
```diff
@@ -0,0 +1,198 @@
+{ 
+    "_disk_gt_70pct_10m": {
```
agilejon: Whoa, does this mean we could just add new conditions here and the appropriate alerting policies would be automatically created?

chrisgilbert: Sure, if we want it to work that way.  I intended to make it easy to see which servers already have a condition and which don't (so we could then create the alert for that condition automatically where required).  

In reality it's been a bit trickier because we are using several different metric types (like cloudwatch, winagent, linux agent) and the JSON objects don't match easily.  But we should be able to make it more consistent, and then add new conditions to the config file and apply them as we like.

tyr/alerts/stackdriver.py
chrisgilbert: Sure, that was mostly for convenience as it started as a standalone script.

tyr/alerts/stackdriver.py
chrisgilbert: I will call r.raise_for_status() here to do that.

chrisgilbert:
I'm considering just merging this, it's feature flagged to be off by default in any case.

chrisgilbert:
Merging this now, it's disabled by default but we may need to start using it soon.

PR #73, opened 2015-08-04
Solr Integration

citruspi:
This adds support for provisioning master and slave Solr nodes. The primary difference is that when provisioning a slave node, you must include a `master` argument for the slave to talk to.

Example:

```
from tyr.servers.solr import SolrMasterNode, SolrSlaveNode

master = SolrMasterNode(group='monolith')
master.autorun()

slave = SolrSlaveNode(group='monolith', master='t-solr-master.thorhudl.com')
slave.autorun()
```


malkolm:
looks good to me

PR #74, opened 2015-08-12
Missing import in stackdriver module

alexdebrie:
Simple fix of a missing import.

Related Q: because the stackdriver utilities module gets imported into `tyr.servers.server.py`, Tyr won't work if you don't have the Stackdriver environment variables set. Is this desired?

### DIFF
--- a/tyr/utilities/stackdriver.py
+++ b/tyr/utilities/stackdriver.py
@@ -3,6 +3,7 @@ import os
 import json
 import time
 import logging
+import sys
 
 log = logging.getLogger('Tyr.Utilities.Stackdriver')
 if not log.handlers:


wliao008:
I think using env vars here is ok? The other option is to read in from a s3 file, or pass in all the parameters at run time, which might be a chore.

alexdebrie:
@wliao008 I'm not opposed to using environment variables, but maybe put checks like [these](https://github.com/hudl/Tyr/blob/b5d0a8125104c6c1a725655e26bce6b1aeea4b0f/tyr/utilities/stackdriver.py#L21) inside a function where they're needed. My problem was that I was trying to spin up a server and it yelled at me that I didn't have those env variables set. I wasn't actually going to be using anything from the stackdriver module, but it gets imported in the base `server.py` and thus requires those env variables to do **anything**.

wliao008:
@alexdebrie ah i see, i thought you were just wondering if that should be used at all. for server spin up, IIRC it was setting the instance to maintenance mode, so it was using stackdriver. But looking thru `server.py`, i only see it used for termination now, so maybe that got changed? If that's the case, like you said, checking when its needed would be the way to go. Maybe submit an issue to have this looked at?

alexdebrie:
That's a good idea. I'll submit one when I get a second. As for the tiny change in this PR, I think it needs done either way as `sys` was missing as an import.

Thanks @wliao008

citruspi:
I'm going to close this since the import statement for `sys` has been added to `tyr/utilities/stackdriver.py`.

PR #75, opened 2015-08-18
Altering policies for SNS to allow add/removal of config subscription

chrisgilbert:
A small alteration to SNS IAM policies to allow listing of config subscriptions. This is for the killingserverssofly script.


tyr/policies/ec2.py
```diff
@@ -211,8 +243,10 @@
         "sns:Subscribe"
       ],
       "Resource": [
-        "arn:aws:sns:us-east-1:761584570493:p-outpost",
-        "arn:aws:sns:us-east-1:761584570493:alyx3"
+        "arn:aws:sns:us-east-1:761584570493:prod-outpost",
```
wliao008: I don't think `prod-outpost` exist?

tyr/policies/ec2.py
```diff
@@ -242,8 +276,10 @@
         "sns:Subscribe"
       ],
       "Resource": [
-        "arn:aws:sns:us-east-1:761584570493:s-outpost",
-        "arn:aws:sns:us-east-1:761584570493:alyx3"
+        "arn:aws:sns:us-east-1:761584570493:stage-outpost",
```
wliao008: `stage-outpost` doesn't exist as well i think

Z3r0Sum: Yeah, curious as to why you changed the topic ARN?  Looks like the following exist for outpost:

```
t-outpost arn:aws:sns:us-east-1:761584570493:t-outpost
p-outpost arn:aws:sns:us-east-1:761584570493:p-outpost
s-outpost arn:aws:sns:us-east-1:761584570493:s-outpost
```

tyr/policies/ec2.py
Z3r0Sum: Can you remove this line?  It doesn't seem like test-outpost exists and you already have line for `t-outpost`.

```
t-outpost arn:aws:sns:us-east-1:761584570493:t-outpost
```

tyr/policies/ec2.py
chrisgilbert: Thanks, I have removed these now.  To be honest I was just worried there might be some using an old name - I should have just checked really.

PR #77, opened 2015-09-02
Apply MongoDB suggested changes

jamiegs:
Increase journal volume speed.
Enable EBS optimized for mongo data nodes.

### DIFF
--- a/tyr/servers/mongo/data.py
+++ b/tyr/servers/mongo/data.py
@@ -77,18 +77,26 @@ class MongoDataNode(MongoReplicaSetMember):
                 {
                     'user': 'mongod',
                     'group': 'mongod',
-                    'size': 10,
-                    'iops': 0,
+                    'size': self.data_volume_size,
+                    'iops': self.data_volume_iops,
+                    'device': '/dev/xvdf',
+                    'mount': '/volr'
+                },
+                {
+                    'user': 'mongod',
+                    'group': 'mongod',
+                    'size': 25,
+                    'iops': 250,
                     'device': '/dev/xvdg',
                     'mount': '/volr/journal'
                 },
                 {
                     'user': 'mongod',
                     'group': 'mongod',
-                    'size': self.data_volume_size,
-                    'iops': self.data_volume_iops,
-                    'device': '/dev/xvdf',
-                    'mount': '/volr'
+                    'size': 50,
+                    'iops': 250,
+                    'device': '/dev/xvdh',
+                    'mount': '/mongologs',
                 }
             ])
 
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -32,6 +32,7 @@ class MongoNode(Server):
     def configure(self):
 
         super(MongoNode, self).configure()
+        self.ebs_optimized = True
 
         if self.environment == "prod":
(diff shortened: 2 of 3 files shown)


tyr/servers/mongo/data.py
```diff
@@ -77,18 +77,26 @@ def bake(self):
...
+                    'size': self.data_volume_size,
+                    'iops': self.data_volume_iops,
+                    'device': '/dev/xvdf',
+                    'mount': '/volr'
+                },
+                {
+                    'user': 'mongod',
+                    'group': 'mongod',
+                    'size': 25,
+                    'iops': 250,
```
citruspi: Do we always want IOPS regardless of the environment?

tyr/servers/mongo/data.py
```diff
@@ -77,18 +77,26 @@ def bake(self):
...
                     'user': 'mongod',
                     'group': 'mongod',
-                    'size': self.data_volume_size,
-                    'iops': self.data_volume_iops,
-                    'device': '/dev/xvdf',
-                    'mount': '/volr'
+                    'size': 50,
+                    'iops': 250,
+                    'device': '/dev/xvdh',
+                    'mount': '/mongologs',
```
citruspi: So, going based on the order of operations in the [`role_mongo::default`][0] recipe, these volumes will be [created and mounted][1] and then the recipe will attempt to [either link an ephemeral volume to the log directory or create a directory for the logs if it can't find an ephemeral volume][2].

I'm curious what happens if the instance has an ephemeral volume and Chef attempts to create a symbolic link with a name of a directory that already exists.

Either way, if we're going to use an EBS volume for `/mongologs` regardless of the environment, should we also remove [these lines][3]?

And would we want to apply [`role_mongo::blockdev_optimizations`][4] to the log volume?

[0]: https://github.com/hudl/role_mongo/blob/master/recipes/default.rb
[1]: https://github.com/hudl/role_mongo/blob/master/recipes/default.rb#L25
[2]: https://github.com/hudl/role_mongo/blob/master/recipes/default.rb#L40
[3]: https://github.com/hudl/role_mongo/blob/master/recipes/default.rb#L40-L53
[4]: https://github.com/hudl/role_mongo/blob/master/recipes/blockdev_optimizations.rb

tyr/servers/mongo/data.py
```diff
@@ -77,18 +77,26 @@ def bake(self):
...
                 },
                 {
                     'user': 'mongod',
                     'group': 'mongod',
-                    'size': self.data_volume_size,
-                    'iops': self.data_volume_iops,
-                    'device': '/dev/xvdf',
-                    'mount': '/volr'
+                    'size': 50,
+                    'iops': 250,
```
citruspi: Do we always want IOPS regardless of the environment? Also, would it make sense to have arguments and variables `log_volume_iops` and `log_volume_size`?

tyr/servers/mongo/data.py
jamiegs: we have some role_mongo changes come to this also

tyr/servers/mongo/data.py
citruspi: Also, would it make sense to have arguments and variables `journal_volume_iops` and `journal_volume_size`?

jamiegs: I think once we figure out the correct value for these we'll just keep it set.

PR #78, opened 2015-09-03
Add configurable values for mongo data node EBS volumes.

Z3r0Sum:
- Both the journal and log EBS data volumes had static values for size
    and iops, which becomes problematic when dealing with multiverse
    mongo instances as they vary quite a bit.
  - Add capability to pass in the above values.

### DIFF
--- a/tyr/servers/mongo/data.py
+++ b/tyr/servers/mongo/data.py
@@ -36,50 +36,50 @@ class MongoDataNode(MongoReplicaSetMember):
         self.log_volume_iops = log_volume_iops
 
     def validate_ebs_volume(self, volume_type):
-        eval_volume_size = 0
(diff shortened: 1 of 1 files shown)


citruspi:
You probably want to [set some default values][0] and [validate the IOPS to size ratio][1].

[0]: https://github.com/hudl/Tyr/blob/struct-add_dynamic_ebs_setup_for_mongo_data_nodes/tyr/servers/mongo/data.py#L47
[1]: https://github.com/hudl/Tyr/blob/struct-add_dynamic_ebs_setup_for_mongo_data_nodes/tyr/servers/mongo/data.py#L72

Z3r0Sum:
Alright, I'll work on that.

PR #80, opened 2015-09-03
Re-Determine the Replica Set Primary After a Replacement Node Finishes Syncing

citruspi:
After a MongoDB replacement node finishes syncing, re-determine the replica set's primary in case it has changed since the sync started.


jamiegs:
:+1:

PR #81, opened 2015-09-07
Fix incorrect resource names for outpost SNS

chrisgilbert:
This is a small fix for the naming of the SNS topics in the policies file.


wliao008:
nice :+1:

chrisgilbert:
Thanks Wei, that was quick :)

PR #83, opened 2015-09-14
forgot to move chef_path when fixing 80 char limit

malkolm:


Z3r0Sum:
Looks like the order was just changed?  I didn't think order mattered when passing name-valued parameters to a python function?  Not sure this will fix the issue.

Z3r0Sum:
This should solve the issue:

```
         super(ElasticsearchServer, self).__init__(group, server_type,
                                                    instance_type, environment,
                                                    ami, region, role, keypair,
                                                    availability_zone,
                                                    security_groups,
                                                    block_devices,
                                                    chef_path,
                                                    subnet_id, dns_zones)
```

PR #85, opened 2015-09-17
Only Enabled EBS Optimization on MongoDB Nodes in Prod

### LINKED ISSUE #84: We should only mandate ebs optimized instances for prod
in /tyr/servers/mongo/node.py, self.ebs_optimized = True, should only be set for nodes where self.environment == "prod"

Currently you cannot provision a non ebs-optimized server in stage or test.

citruspi:
Fixes #84 

> in /tyr/servers/mongo/node.py, self.ebs_optimized = True, should only be set for nodes where self.environment == "prod"

> Currently you cannot provision a non ebs-optimized server in stage or test.

### DIFF
--- a/tyr/servers/mongo/node.py
+++ b/tyr/servers/mongo/node.py
@@ -32,7 +32,9 @@ class MongoNode(Server):
     def configure(self):
 
         super(MongoNode, self).configure()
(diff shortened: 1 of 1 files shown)


malkolm:
:+1:  thank you

PR #86, opened 2015-09-18
Further fix for INF-952 permissions

chrisgilbert:
This is another change to the IAM role permissions in order to allow web servers to send the autoscale lifecycle completion message.


wliao008:
lgtm :shipit:

PR #87, opened 2015-09-23
Add mongo cm monitoring type

Z3r0Sum:
- There was a cookbook developed for automatically configuring the
  Mongo Monitoring Agent instances in an attempt to easily move them
  to the VPC. The goal of this new Server Type is to leverage that
  cookbook and make it easy to spin up this server type in the future.


jamiegs:
Cloud Manager monitors were moved to docker.. this PR is no longer needed.

PR #90, opened 2015-09-25
Fixing bug with overriding instance type

chrisgilbert:
Fix a bug that prevented overriding the instance type on a scripts server.


Z3r0Sum:
:+1:

PR #91, opened 2015-10-16
NO-TICK: Increase ASG grace period to 900

wliao008:
There have been several incidents where ASG spins up an instance, it goes on the vicious cycle of `start -> status check failed -> terminate -> start -> ..` Currently a web server takes roughly 20 minutes to spin up, so increasing the grace period to 900 has so far mitigated this issue.


malkolm:
Looks good

PR #94, opened 2015-12-04
Consolidated the SolrMasterNode and SolrSlaveNode classes into SolrNode class

citruspi:
Instead of maintaining the same logic in `SolrMasterNode` and `SolrSlaveNode`, I consolidated them into a `SolrNode` class which they can just subclass.

### DIFF
--- a/tyr/servers/solr/node.py
+++ b/tyr/servers/solr/node.py
@@ -1,0 +1,72 @@
+from tyr.servers.server import Server
+
+
+class SolrNode(Server):
+
+    SERVER_TYPE = 'solr'
+    CHEF_RUNLIST = ['role[RoleSolr]']
+    IAM_ROLE_POLICIES = [
+        'allow-volume-control',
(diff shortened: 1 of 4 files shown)


tyr/servers/solr/node.py
```diff
@@ -0,0 +1,72 @@
...
+    IAM_ROLE_POLICIES = [
+        'allow-volume-control',
+        'allow-get-solr-schema'
+    ]
+
+    def __init__(self, group=None, server_type=None, instance_type=None,
+                 environment=None, ami=None, region=None, role=None,
+                 keypair=None, availability_zone=None, security_groups=None,
+                 block_devices=None, chef_path=None, subnet_id=None,
+                 dns_zones=None, is_master=True, is_slave=False,
```
jamiegs: do we really need a is_master and is_slave? There's no 3rd option is there? If is_master = false isn't it a slave?

PR #95, opened 2015-12-15
Specifying the instance type in the launch configuration, allowing th…

llaughlin:
Before, the specified `instance_type` wasn't getting respected when spinning up new servers. This was because it wasn't getting used when creating launch configurations.

This PR passes the `instance_type` down through the `Autoscaler` to the `LaunchConfiguration` constructor, which fixed a bug causing new servers spun up with Tyr to all be `m1.small`.

### DIFF
--- a/tyr/clusters/autoscaling.py
+++ b/tyr/clusters/autoscaling.py
@@ -10,7 +10,6 @@ class AutoScaler(object):
     '''
     def __init__(self, launch_configuration,
                  autoscaling_group,
-                 instance_type,
                  node_obj,
                  desired_capacity=1,
                  max_size=1,
@@ -24,7 +23,6 @@ class AutoScaler(object):
 
         self.launch_configuration = launch_configuration
         self.autoscaling_group = autoscaling_group
-        self.instance_type = instance_type
         self.desired_capacity = desired_capacity
 
         # You can set a list of availability zones explicitly, else it will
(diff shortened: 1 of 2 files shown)


tyr/clusters/autoscaling.py
```diff
@@ -10,6 +10,7 @@ class AutoScaler(object):
     '''
     def __init__(self, launch_configuration,
                  autoscaling_group,
+                 instance_type,
```
chrisgilbert: I don't think this line is actually needed - it seems the node_obj passes this setting anyway - the problem was just in the launch configuration constructor.

tyr/clusters/autoscaling.py
```diff
@@ -77,6 +79,7 @@ def create_launch_configuration(self):
                                      get_security_group_ids(
                                          self.node_obj.security_groups),
                                      user_data=self.node_obj.user_data,
+                                     instance_type=self.instance_type,
```
chrisgilbert: Ditto - the node_obj contains the instance_type

chrisgilbert: So this would say instance_type=self.node_obj.instance_type

tyr/clusters/iis.py
```diff
@@ -86,6 +86,7 @@ def provision(self):
         self.log.info('Creating autoscaler')
         auto = AutoScaler(launch_configuration=self.launch_configuration,
                           autoscaling_group=self.autoscaling_group,
+                          instance_type=self.instance_type,
```
chrisgilbert: This is also not needed then I think.

tyr/clusters/autoscaling.py
```diff
@@ -23,6 +24,7 @@ def __init__(self, launch_configuration,
 
         self.launch_configuration = launch_configuration
         self.autoscaling_group = autoscaling_group
+        self.instance_type = instance_type
```
chrisgilbert: This can go too - think it's just the one line that needs changing in this file

chrisgilbert:
Looks good, gonna merge.  Thanks for your fix @llaughlin :+1:

PR #96, opened 2016-01-20
Micro add windows user data

Z3r0Sum:
- As we start using chef on our windows servers, this will serve as a common bootstrap user data script to be run in order to prep the server for a chef run.
- It relies on the following tags being set on the instance: Environment and Role.


scripts/user_data_base.ps1
```diff
@@ -0,0 +1,163 @@
...
+$chef_tag_role = Get-EC2Tag -Filters @( @{Name="key"; values="Role"}, @{Name="resource-id"; Values=$instanceId})
+$chef_environment = Get-EC2Tag -Filters @( @{Name="key"; values="Environment"}, @{Name="resource-id"; Values=$instanceId})
+$windows_version = "2012r2"
+$chef_version = "chef-client-12.6.0-1-x86.msi"
+$server_name = $servername.Trim()
+$chef_env = $chef_environment.Value.ToLower().Trim()
+$client_rb = @"
+base_dir = "C:/Chef"
+log_level        :info
+log_location     STDOUT
```
jamiegs: should we set this so all future runs of chef get logged somewhere?

scripts/user_data_base.ps1
```diff
@@ -0,0 +1,163 @@
+<powershell>
+#User Data script
+$ErrorActionPreference = "Stop";
+
+Import-Module "C:\Program Files (x86)\AWS Tools\PowerShell\AWSPowerShell\AWSPowerShell.psd1"
```
jamiegs: I could be wrong but I don't think this is actually needed anymore, I think it's automatically loaded now.

scripts/user_data_base.ps1
Z3r0Sum: @jamiegs We could set it there - the chef-client cookbook automatically creates a default log location and log file in C:\chef\log

jamiegs:
@Z3r0Sum Think this is ok to have merged in?

Z3r0Sum:
@jamiegs I think so.  @citruspi You ok with the idea behind this?

citruspi:
So this is the user data that would be set on launch for Windows instances in AWS?

If that's the case, I don't think `/scripts` is the right place for it to go.

For some background, in a Python project, command line scripts may be included to provide access to the package from the command line (as opposed to having to open a Python interpreter, import the package, and then interact with it programmatically or having to write a script to use the package).

For example, Tyr provides the module `compact_mongo_collection` in the `tyr.utlities` package for compacting MongoDB servers.

A user could use this programmatically:

``` sh
$ python
>>> from tyr.utilities.compact_mongo_collection import compact_mongodb_server
>>> hosts = [
...    'p-monolith-mongo-rs1-use1c-01.app.hudl.com',
...    'p-monolith-mongo-rs2-use1c-01.app.hudl.com',
...    'p-monolith-mongo-rs3-use1c-01.app.hudl.com',
...    'p-monolith-mongo-rs4-use1c-01.app.hudl.com',
...    'p-monolith-mongo-rs5-use1c-01.app.hudl.com',
... ] 
>>> compact_script_version = '2012-07-22'
>>> for host in hosts:
...    compact_mongodb_server(host, compact_script_version)
```

However, we also provide a script (`/scripts/compact-mongodb-servers`) which allows them to use the functionality from the command line by providing a JSON or YAML file with a list of servers to compact.

When the Python package is installed on a machine, the scripts which are specified in `setup.py` are placed in the user's `$PATH`, specifically in the Python installation's `/bin` directory. (If using a virtual environment - _and you should be :suspect:_ - then the script will be placed in `/path-to-virtualenv/bin/`.)

So now it's as simple as

``` sh
$ compact-mongodb-servers --source hosts.yaml
```

(This is what happens behind the scenes when replacing MongoDB servers on `p-infra-comm`.)

A simple test - is the code being used to provide a command line interface **to** the package or is it being used **by** the package?

If the answer is the latter or neither, then it doesn't belong in `/scripts`.

TL;DR If the plan is for Tyr to set this user data when launching Windows instances, `/scripts` is not the correct place for it.

citruspi:
Thinking more about this, if the user data script is not placed in `/scripts`, I think there's three ways to go about it.

1. A multi-line string in the code for setting the user data.

    So basically we'd add another variable similar to [the user data for Linux instances][0].

    However, at 372 lines, that string would be 26% of the code in the `tyr.servers.server` module, not to mention the mess that would be editing it.

2. Read the script from S3.

    Instead of having the script live in the Tyr repository, we'd move it to another repository and configure TeamCity to push the script to S3. Tyr would read the script if a Windows instance is being launched and would populate the user data that way.

    However, that means keeping data required by Tyr in a separate repository.

3. Make use of `data_files` files in `setup.py`.

    We can keep the script in another top level directory and instruct `setup.py` to treat it as a data file. When Tyr is installed, the file would be copied to the specified directory and then Tyr would traverse its installation path to access it.

    

[0]: https://github.com/hudl/Tyr/blob/master/tyr/servers/server.py#L403

citruspi:
So, going a little further with option three, I just tested out a quick little implementation of what it might look like.

Let's say I have a Python package called `pkg`, with which I'd like to include some sort of file for use by the package.

All my code for the package is in `/pkg`, my data file is in `/data`, and we have good old `setup.py`.

``` sh
.
├── data
│   └── hello.txt
├── pkg
│   └── __init__.py
└── setup.py
```

Here's what each file looks like:

``` sh
# data/hello.txt

Hello World!
```

``` python
# pkg/__init__.py

import os


def get_data(f):
    root = os.path.abspath(os.path.dirname(__file__))
    path = os.path.join(root, '..', 'data', f)

    return open(path, 'r')
```

``` python
# setup.py

from setuptools import setup

setup(
    name='pkg',
    version='0.0.1',
    packages=['pkg'],
    zip_safe=False,
    include_package_data=True,
    platforms='any',
    data_files=[
        ('data', ['data/hello.txt'])
    ]
)
```

If I install this package, I can now do:

``` sh
$ python
>>> from pkg import get_data
>>> data = get_data('hello.txt')
>>> data.read()
'Hello World!\n'
```

citruspi:
If we decide on option three, I've implemented support for data files to Tyr in #99. So you'd be able to access the user data file with:

``` python
from tyr.helpers import data_file

try:
    f = data_file('user_data_base.ps1')
    user_data = f.read()
except IOError:
    # Handle error reading file
    pass
```

Z3r0Sum:
@citruspi I like option 3 :+1:

citruspi:
Great, once #99 is merged in you should be good to go.

Side note, once #99 is merged in, at some point it might make sense to move the Linux user data that is in `tyr.servers.server` to a data file like `linux_user_data.sh`, if only to make it easier to edit and to avoid having long, unnecessary blobs of text in there.

Related, if possible, could we include `windows` in the file name for the user data? Something like `windows_user_data.ps1`. That way we're not relying on the file extension to differentiate between user data for different AMIs.

Z3r0Sum:
Yeah, that makes sense - I'll move it to `windows_user_data.ps1`.

jamiegs:
@Z3r0Sum it's not directly a powershell file.  I feel if we rename it to a *.ps1 file we should remove the <powershell> tags and have those get inserted in code.

Z3r0Sum:
@jamiegs True, how about just `windows_user_data` for now?

citruspi:
We probably want to use the file extension that is appropriate for the format.

So if we remove the `<powershell>` tags, it would be a Powershell script and `ps1` would be the appropriate extension.

If we keep the `<powershell>` tags, it would be an XML file and `xml` would be the appropriate extension.

jamiegs:
Well, technically, I don't think the file format really follows the XML format either.  The way we're using it it does.. we only use 1 tag within it, but you are able to use multiple.. and they all based off of the root.. which XML only allows for 1 root element.

citruspi:
Interesting. 

So is there any case for Windows user data where we'd want to use multiple root tags? If so, we'd probably want to leave the tags in the script so that Tyr doesn't need to figure out where to place multiple tags.

And if that's the case, any thoughts on what the appropriate extension would be?

If we only plan on using a single root tag for now, I'd be comfortable removing the tag from the script and leaving the extension `ps1`. Whenever Tyr detects Powershell being used for user data it could just wrap the whole document in the tag.

For what it's worth, I like having the tag(s) as part of the initial document because then Tyr doesn't need to think about how to format the user data before using it. It reads the file and then uses the contents, no logic involved.

But I also like using the `ps1` extension because it would provide accurate syntax highlighting for the script, which would be useful when making changes.

citruspi:
@jamiegs.

After some discussion with @Z3r0Sum, I'm not a huge fan of the idea of modifying user data files in Tyr, apart from rendering templates. I think that what you put in should be, for the most part, what you get out.

As for the file extension, I'm fine with `xml`, or `ps1`, but I'm not a fan of not having an extension.

Either way, I'd like to get the ball rolling on this. Thoughts?

citruspi:
To clarify, I'm open to other extensions, but I think that `xml` and `ps1` are the most accurate, even if they're not 100% correct.

I considered `ps`, but that's normally used for PostScript files.

jamiegs:
@citrispi.  if you'd want to use templates then, how about a powershellWrapper template, that's 
```<powershell>
{{powershellContent}}
</powershell>```

then have a user-data.ps1 file that gets read into the {{powershellContent}}
sorta jank, but then uses correct file formats for the extensions.

PR #97, opened 2016-02-03
MICRO  - add repo_upgrade disable

jamiegs:
Adds change to userdata to disable chef repo upgrading to newer amazon linux version.


citruspi:
Looks good to me. :+1:

PR #99, opened 2016-02-29
Support for Data Files

citruspi:
If we decide to use `data_files` in #96, this should make it pretty simple.

To add a data file, simply place it in the `data` directory and add it to `data_files` in `setup.py`.

The entry in `data_files` would look something like this:

``` python
('data', ['data/<file_name>'])
```

So for the data file `user_data_base.ps1`, `setup.py` might look like this:

``` python
from setuptools import setup

setup(
    ...
    platforms='any',
    data_files=[
        ('data', ['data/user_data_base.ps1'])
    ]
)
```

To access the data file `user_data_base.ps1` from within Tyr, it'd look like this:

``` python
from tyr.helpers import data_file

try:
    f = data_file('user_data_base.ps1')
    user_data = f.read()
except IOError:
    # Handle error reading file
    pass
```

### DIFF
--- a/data/.placeholder
+++ b/data/.placeholder
@@ -1,0 +1,3 @@
+Just a placeholder file until something is added to the directory.
+
+So, y'know, the directory doesn't disappear from git.
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ setup(
     version='0.0.1',
     author='Mihir Singh (@citruspi)',
     author_email='mihir.singh@hudl.com',
+    test_suite='nose.collector',
     packages=['tyr'],
     zip_safe=False,
     include_package_data=True,
@@ -26,7 +27,8 @@ setup(
         'paramiko',
         'click',
         'PyYAML',
-        'requests'
+        'requests',
+        'nose'
     ],
     scripts=[
         'scripts/replace-mongodb-servers',
--- a/setup.py
+++ b/setup.py
@@ -20,5 +20,8 @@ setup(
     scripts=[
         'scripts/replace-mongodb-servers',
         'scripts/compact-mongodb-servers'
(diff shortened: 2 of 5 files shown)


citruspi:
I've added support for automatically loading the data files in `setup.py`, so now using data files is even easier!

Just place the file in the `data` directory and it'll automatically be included when the project is installed.

On a related note, I've created the `data` directory and included a placeholder file so that `setup.py` doesn't have a seizure when it doesn't find the directory since git doesn't let you add empty directories.

The file can be removed whenever an actual data file is added.

citruspi:
It took 381 days to get here, but we now have tests. Well, two of them.

More to come in the future.

Z3r0Sum:
@citruspi Cool - Do you plan on merging this in?  I'll include this in the other PR once it is.  cc: @jamiegs

citruspi:
I do. It's not a lot of code, but I figured I'd get another set of eyes on it before merging.

Z3r0Sum:
@citruspi Curious, doesn't seem like you're using `unittest`?  Seems like a fairly nice module that includes `setUp` and `tearDown` functions.

```
import unittest

class WidgetTestCase(unittest.TestCase):
    def setUp(self):
        self.widget = Widget('The widget')

    def tearDown(self):
        self.widget.dispose()
        self.widget = None

    def test_default_size(self):
        self.assertEqual(self.widget.size(), (50,50),
                         'incorrect default size')

    def test_resize(self):
        self.widget.resize(100,150)
        self.assertEqual(self.widget.size(), (100,150),
                         'wrong size after resize')
```

Seems like a nice framework that would create uniform test creation in the future.  Thoughts?

PR #100, opened 2016-03-18
MICRO-200 - Added the option to set disk size and iops.

acsportstec:
- Output the connection error for SSH rather than generic message.
- Changed the AMI to the latest version. 
- Added set drive size & iops for Solr servers.


jamiegs:
looks good :+1:

PR #101, opened 2016-03-21
Update README.md

atharvai:
added missing dependency on setting `STACKDRIVER_` environment variables and requirement of chef creds


chrisgilbert:
Thanks for this @atharvai - LGTM :+1:

PR #102, opened 2016-04-06
Use A Records with Private IP Addresses in Place of CNAMES with Private Hostnames

citruspi:
**Issue**:
While provisioning MongoDB nodes for use with MongoDB Cloud Manager yesterday, @Z3r0Sum and I noticed that the nodes would come up within Cloud Manager using their private hostnames (e.g. `ip-165-23-127-185.ap-southeast-2.compute.internal`) instead of their public names (e.g. `p-dvms-mongo-rs1-use1c-01...`).

While not a huge issue, it makes it harder to find servers in Cloud Manager and will likely make future automation with the Cloud Manager API more difficult.

**Solution**:
Using `A` records with the private IP address instead of `CNAME` records with the private hostnames fixes this issue.

**Relevant GIF**:
giphy

### DIFF
--- a/tyr/servers/server.py
+++ b/tyr/servers/server.py
@@ -259,9 +259,9 @@ class Server(object):
                             'ttl': 60
                         },
                         {
-                            'type': 'CNAME',
+                            'type': 'A',
                             'name': '{hostname}.',
-                            'value': '{private_dns_name}',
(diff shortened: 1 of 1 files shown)


Z3r0Sum:
👍

PR #103, opened 2016-04-14
Include an Additional EBS Volume for Swap on MongoDB Instances that Lack Instance Storage

citruspi:
To quote myself from [`role_mongo` pull request 57][0]:

> EC2 provides many different types of instances, some which support instance storage (ephemeral storage), and some which don't.

> On instances that support instance storage, the device (`/dev/xvdc`) is formatted as `ext3` and mounted at `/media/ephemeral0`. The swap file, created by `role_mongo`, is then placed there. On instances that don't support instance storage, the swap file is still placed there, but it's now on the root device. This can cause problems when the root device is ~8 GB and holds the operating system as well as a swap file that's nearly 50% of it's size.

> A separate pull request to Tyr fixes that - MongoDB instances that don't support instance storage will automatically be launched with an additional 8 GB EBS volume located at `/dev/xvdc` and mounted to `/media/ephemeral0`.

This is that pull request.

[0]: https://github.com/hudl/role_mongo/pull/57

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ setup(
         'click',
         'PyYAML',
         'requests',
-        'nose'
+        'nose',
+        'cloudspecs'
     ],
     scripts=[
         'scripts/replace-mongodb-servers',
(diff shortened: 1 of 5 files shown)


Z3r0Sum:
:+1:

PR #104, opened 2016-04-26
Move build multiverse script

jamiegs:
Moves the build multiverse script into the right place and adds it to the setup.py so it to get installed.


citruspi:
👍

PR #110, opened 2016-05-27
Alpha chefd web server provision

acsportstec:
For the ASG with Tyr you can now:

* add parameters for mongos_preference, mongos_version, mongos_servers, which will get set as tags on the ASG
* Checks against AMI if it is not available then the program exits with a error, updated to the latest 
* Uses managed policies not Inline policies
* make the root volume size configurable, default to 40 gig
* chef-node security group added 
* Tyr now uses the file 'user-data' to pull in the user-data for the ASG


scripts/build-mv-service
```diff
@@ -18,21 +18,24 @@ if not log.handlers:
 
 
 def build_mv_service(
-    service=None,
+    service="supportinvestigator",
```
jamiegs: These need to be set back to what they were.

jamiegs:
Where are the changes for the managed policies?

.gitignore
```diff
@@ -3,3 +3,4 @@
 /dist
 *.pyc
 .ropeproject
+*.swp
```
jamiegs: nice!

jamiegs:
There's also a lot of indentation issues.

jamiegs:
why'd this get closed?

PR #111, opened 2016-05-31
Alpha chefd web server provision

acsportstec:
- You can now specify tags & mongos tags for the ASG (mongos_preference & mongos_servers)
- Get the most recent windows 2012 r2 base AMI (if none has been provided)
- add the new managed policies to the IAM role rather than using the inline policies, taking into account environment (stage / prod)
- Root volume size configurable, default to 40 gig
- Chef-nodes included in security group
- Removed p-web security group
- change the user-data on the launch config to read in user-data file in script


scripts/build-mv-service
```diff
@@ -49,13 +52,19 @@ def build_mv_service(
     prod_environment = 'prod'
     stage_environment = 'stage'
 
+    tags = list()
+    tags.append({'name':'chef:hudl:mongos_preference', 'value':mongos_preference})
+    tags.append({'name':'chef:hudl:mongo_config_servers', 'value':mongos_servers})
```
jamiegs: what about mongos_version?

tyr/servers/server.py
```diff
@@ -63,6 +65,23 @@ def __init__(self, group=None, server_type=None, instance_type=None,
         self.add_route53_dns = add_route53_dns
         self.ebs_optimized = False
 
+
+    def get_latest_ami(self, ami=None, filter=None):
+        if ami is not None:
+            return ami
+
+        if filter is None:
+            ami_filter = { 'architecture': 'x86_64', 'name':'Windows_Server-2012-R2_RTM-English-64Bit-Base-*' }
```
jamiegs: what about if it's a linux server?

tyr/servers/server.py
```diff
@@ -597,74 +617,89 @@ def resolve_iam_role(self):
...
 
-                if rolePolicy is None:
-                    self.log.info("No policy defined for {policy}".format(
-                                  policy=policy))
-                    continue  # Go to the next policy
+            for m_policy in self.IAM_MANAGED_POLICIES:
+                #self.log.info("Attaching {m_policy}".format(m_policy=m_policy))
+                m_policy_id = m_policy.format(environment=self.environment)
+                arn = self.iam.get_user().user.arn
+                ARN = arn[arn.find('::')+2:arn.rfind(':')]
```
jamiegs: seems like this is suppose to be the account_id, not the ARN.. variable name is confusing.

tyr/servers/server.py
```diff
@@ -597,74 +617,89 @@ def resolve_iam_role(self):
...
 
-                        self.log.info('Added policy "{policy}"'.format(
+                        self.log.warn('Policy "{policy}" has been modified'.format(
                                       policy=policy))
-                    except Exception as e:
-                        self.log.error(str(e))
-                        raise e
+
+                        try:
+                            self.iam.delete_role_policy(self.role, policy)
```
jamiegs: so you're deleting the policy then adding it again?  That would cause a brief moment when that policy is removed from a server if there's any using that IAM role.  This could potentially cause production issues.

jamiegs:
There's lots of indentation issues.  If you're having problems with the IDE using the wrong thing (tabs vs spaces) a way to fix those without messing with the IDE is editing the files on the github web interface

jamiegs:
LGTM

tyr/servers/iis/node.py
```diff
@@ -69,31 +69,23 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
-            "key": "{env}-mv-web/init.config.json".format(
-                env=self.environment[0]),
-            "mongos": self.mongos_service,
-            "mongoServers": self.mongo_servers
-        }
-
-        ud = json.dumps(data)
-        self.log.info("Setting user data to: {ud}".format(ud=ud))
+        # read in userdata file
+        self.log.info("Setting user data from location [" + os.path.realpath(sys.path[0]) + "/user-data" + "]")
```
jamiegs: This is already in the repo at `/data/user_data_base.ps2`

tyr/servers/iis/node.py
```diff
@@ -69,31 +69,27 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
-            "mongos": self.mongos_service,
-            "mongoServers": self.mongo_servers
-        }
-
-        ud = json.dumps(data)
-        self.log.info("Setting user data to: {ud}".format(ud=ud))
-        return ud
+        # read in userdata file
+        user_data = None
+        try:
```
Z3r0Sum: I think we should handle this better - if we can't find the userdata script for some reason, we should probably exit before creating the LC, ASG, and the subsequent servers as it's a vital aspect of it.

tyr/servers/iis/node.py
```diff
@@ -69,31 +69,27 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
+            f = data_file('user_data_base.ps1')
```
Z3r0Sum: This is named incorrectly as well, should be: `'user_data_base.ps2'`

tyr/servers/server.py
```diff
@@ -597,74 +620,88 @@ def resolve_iam_role(self):
...
 
-                if rolePolicy is None:
-                    self.log.info("No policy defined for {policy}".format(
-                                  policy=policy))
-                    continue  # Go to the next policy
+            for m_policy in self.IAM_MANAGED_POLICIES:
+                m_policy_id = m_policy.format(environment=self.environment)
+                arn = self.iam.get_user().user.arn
+                account_id = arn[arn.find('::')+2:arn.rfind(':')]
+                m_policy_arn = self.iam.get_policy("arn:aws:iam::{account_id}:policy/{policy}".format(account_id=account_id,policy=m_policy_id))
```
citruspi: `m_policy_arn` is never used after being set.

tyr/servers/server.py
```diff
@@ -136,8 +158,9 @@ def configure(self):
         self.establish_route53_connection()
 
         if self.ami is None:
-            self.log.warn('No AMI provided')
-            self.ami = 'ami-8fcee4e5'
+            self.log.warn('No AMI provided, searching for latest one...')
+            self.ami = self.get_latest_ami(self.ami)
```
citruspi: While the idea of retrieving the latest AMI for a platform if one isn't specified sounds nice, we could run into issues, at least on the Linux side.

Amazon Linux AMI updates are also Amazon Linux OS updates and we've had previous OS updates which broke our provisioning process. It would be unfortunate if a new OS update came out and we automatically started using that without vetting that everything will continue to work without issues.

Feel free to leave the implementation of `get_latest_ami`, but I'd prefer to hardcode default AMI IDs for now, instead of automatically using the latest. It'd be useful to have a default AMI per-platform.

tyr/servers/server.py
```diff
@@ -597,74 +620,88 @@ def resolve_iam_role(self):
...
+                m_policy_id = m_policy.format(environment=self.environment)
+                arn = self.iam.get_user().user.arn
+                account_id = arn[arn.find('::')+2:arn.rfind(':')]
+                m_policy_arn = self.iam.get_policy("arn:aws:iam::{account_id}:policy/{policy}".format(account_id=account_id,policy=m_policy_id))
+                self.iam.attach_role_policy("arn:aws:iam::{account_id}:policy/{policy}".format(account_id=account_id, policy=m_policy_id), self.role)
 
-                self.log.info('Policy "{policy}" does not exist'.format(
-                              policy=policy))
+        # Use inline roles instead of managed policies:
+        else:
```
citruspi: IAM roles can support inline policies _and_ managed policies.

tyr/servers/iis/node.py
```diff
@@ -69,31 +69,29 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
+        # read in userdata file
+        user_data = None
+        try:
+            f = data_file('user_data_base.ps2')
+            user_data = f.read()
+        except IOError:
+            # Handle error reading file
+            self.log.critical('Could not load user_data_base, quitting')
+            exit(1)
+            pass
```
citruspi: If we're going to handle the exception, there's no need to use `pass`.

tyr/servers/iis/node.py
```diff
@@ -10,21 +10,23 @@ class IISNode(Server):
...
+    
     IAM_ROLE_POLICIES = [
-        'allow-describe-instances',
-        'allow-create-tags',
-        'allow-describe-tags',
-        'allow-describe-elbs',
-        'allow-set-cloudwatch-alarms',
-        'allow-remove-cloudwatch-alarms',
-        'allow-deploy-web-updates',
+
```
citruspi: An empty list would just be `[]`.

tyr/servers/iis/node.py
```diff
@@ -10,21 +10,23 @@ class IISNode(Server):
...
     ]
 
+
```
citruspi: Two extra new lines.

scripts/build-mv-service
```diff
@@ -49,13 +53,22 @@ def build_mv_service(
     prod_environment = 'prod'
     stage_environment = 'stage'
 
+    tags = list()
```
citruspi: An empty list should be `[]`, but you could probably just make it

```
tags = [{'name':'chef:hudl:mongos_preference', 'value':mongos_preference}]
```

and remove the next line.

tyr/clusters/autoscaling.py
```diff
@@ -90,7 +109,14 @@ def create_autoscaling_group(self):
             self.log.info("Creating new autoscaling group: {g}"
                           .format(g=self.autoscaling_group))
 
+	    # Convert our tags list into something that AWS can understand:
+	    aws_tags = list()
```
citruspi: An empty list would just be `[]`, but you could probably replace the whole loop with

```
aws_tags = [Tag(resource_id=self.autoscaling_group, key=tag['name'],
                value=tag['value'], propagate_at_launch=True) for tag in self.tags]
```

tyr/clusters/autoscaling.py
```diff
@@ -90,7 +109,14 @@ def create_autoscaling_group(self):
...
+	    for tag in self.tags:
+		self.log.info("Adding tag [" + str(tag['name']) + "] with value [" + str(tag['value']) + "]")
+		aws_tags.append(Tag(resource_id=self.autoscaling_group, key=tag['name'], value=tag['value'], propagate_at_launch=True))
+
```
citruspi: Not just an issue on this line, but I've noticed a couple places in this PR where tabs are being used.

FWIW, spaces are preferred in Python and, while some people use tabs, let's not mix the two indentation styles.

scripts/build-mv-service
```diff
@@ -49,13 +53,22 @@ def build_mv_service(
...
+    tags.append({'name':'chef:hudl:mongos_preference', 'value':mongos_preference})
+    if mongos_servers is not '' and mongos_servers is not None:
+      tags.append({'name':'chef:hudl:mongo_config_servers', 'value':mongos_servers})
```
citruspi: Indentation should be four spaces.

tyr/servers/server.py
acsportstec: Good points. It was put in as AWS kill of AMIs, you can bypass this code by hardcoding the AMI as a parameter. Ill leave it to the group to decide if this is still wanted or not (personally I agree that this should probably be removed).

tyr/servers/server.py
acsportstec: Hey mate, my understanding of the Jira was the server should either be setup with managed policies or inline policies but not both, I can correct this if I am not right on this

PR #112, opened 2016-06-13
Create Nginx ASG cluster using Tyr

acsportstec:
Currently our nginx servers are not using auto-scaling in any capacity. This project is to get them in an ASG and scaling automatically.

You can use this script to generate the ASG for the Nginx cluster. (build-nginx-service) as part of change [ALPHA-246](https://jira.hudl.com/browse/ALPHA-246)

### DIFF
--- a/scripts/build-nginx-service
+++ b/scripts/build-nginx-service
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+from tyr.clusters.nginx import NginxCluster
+import logging
+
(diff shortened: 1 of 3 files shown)


jamiegs:
We'll need to change the user-data that gets set on the ASG to what's being used on the craft servers. the current user-data that gets used needs the server to exist on chef already.

jamiegs:
And when it's in an ASG it won't exist.

PR #113, opened 2016-06-27
Alpha dynamic nginx asg

acsportstec:
Changes have been made to Tyr to allow for the Dynamic Autoscaling of the nginx servers. I.e. if the load of the servers start to increase then Amazons ASG should scale automatically to handle that load

You can make your changes in scripts/build-nginx-service and data/user_data_base.ps2, add the appropriate settings for your service before running the build-nginx-service scripts:

Ticket: [ALPHA-246](https://jira.hudl.com/browse/ALPHA-246)


data/user_data_base.ps2
```diff
@@ -1,372 +1,135 @@
...
-catch [Exception]{
-    Write-Output "Failed to run Chef!" | Out-File -FilePath $USERDATA_LOG 
-    Write-Output "Failed to run chef"
-    Write-Output $_.Exception.Message | Out-File -FilePath $USERDATA_LOG -ErrorAction Stop
-    Write-Output $_.Exception.Message
-}
-
-</powershell>
\ No newline at end of file
+Content-Type: multipart/mixed; boundary="===============0035287898381899620=="
```
citruspi: I don't think this is Powershell :)

scripts/build-nginx-service
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
+
+from tyr.clusters.nginx import NginxCluster
+from tyr.helpers import data_file
```
citruspi: `data_file` is never used.

tyr/clusters/nginx.py
```diff
@@ -0,0 +1,113 @@
...
+        self.desired_capacity = desired_capacity
+        self.max_size = max_size
+        self.min_size = min_size
+        self.default_cooldown = default_cooldown
+        self.availability_zones = availability_zones
+        self.health_check_grace_period = health_check_grace_period
+        self.launch_configuration = launch_configuration
+
+        if 'Role' not in self.tags:
+          self.tags.append({'name':'Role', 'value':'RoleNginx'})
```
citruspi: If `self.tags = tags = None`, you're gonna get an `AttributeError`.

tyr/servers/nginx/server.py
```diff
@@ -35,5 +38,19 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
 
     def configure(self):
-
+        super(NginxServer, self).establish_logger()
         super(NginxServer, self).configure()
+
+    @property
+    def user_data(self):    
+        # read in userdata file
+        self.log.info("Loading user-data [user_data_file.ps2]")
```
citruspi: This should probably be moved to something "autoscaling Nginx"-specific that doesn't use `ps2` as the extension.

tyr/servers/nginx/server.py
```diff
@@ -35,5 +38,19 @@ def __init__(self, group=None, server_type=None, instance_type=None,
                                           chef_path, subnet_id, dns_zones)
 
     def configure(self):
-
+        super(NginxServer, self).establish_logger()
```
citruspi: If `establish_logger` isn't being called automatically through your usage of `NginxServer` (as is the case with the `NginxCluster` above) you should probably call `establish_logger` after creating the server instance instead of adding it to `configure`.

tyr/clusters/nginx.py
acsportstec: Role should always be added to tags even if tags = None (tags always at least contains 'Role' of 'RoleNginx').

tyr/servers/nginx/server.py
acsportstec: Sure thing, I just saw that this was how it was called from iis/node.py and followed the same line

tyr/clusters/nginx.py
acsportstec: Ignore me, I see what you mean now, I have added a check for this now

tyr/policies/ec2.py
```diff
@@ -399,3 +398,19 @@
         }
     ]
 }"""
+
+allow_lifecycle_nginx_stage = """{
```
jamiegs: Aren't we going to need one of these for prod too?

data/user_data_chef_provision
```diff
@@ -0,0 +1,135 @@
...
+
+# set the ec2 ohai hints so ohai knows it's in a ec2, VPCs mess it up.
+mkdir -p /etc/chef/ohai/hints
+touch /etc/chef/ohai/hints/ec2.json
+
+# Write first boot json
+echo "{\"run_list\": [\"role[$ROLE]\"]}" > /etc/chef/firstrun.json
+
+# Write Validation key to server
+echo '-----BEGIN RSA PRIVATE KEY-----
```
jamiegs: This is a public repo, we should remove this RSA Key. 
@citruspi how tough would it be to get it removed from github history too?

citruspi: @jamiegs I've removed his key pair from the GitHub history, but it may have already been cached elsewhere, so I've also created a new key pair for him. The one that was in the history shouldn't work anymore, even if it exists elsewhere.

jamiegs:
LGTM

tyr/servers/server.py
```diff
@@ -140,8 +163,9 @@ def configure(self):
         self.establish_route53_connection()
 
         if self.ami is None:
-            self.log.warn('No AMI provided')
-            self.ami = 'ami-8fcee4e5'
```
citruspi: This should still be the default fall back. The default value of `self.ami` is `None`, removing this as the fall back will require developers to either always provide this AMI or to  always use `get_latest_ami`.

tyr/servers/server.py
```diff
@@ -65,7 +69,26 @@ def __init__(self, group=None, server_type=None, instance_type=None,
         self.classic_link = classic_link
         self.add_route53_dns = add_route53_dns
         self.ebs_optimized = False
+        self.platform = platform
         self.create_alerts = False
+        self.use_latest_ami = use_latest_ami
+
+    def get_latest_ami(self, ami=None, platform="linux"):
+        if self.ami is not None or self.use_latest_ami is False:
+            self.log.info('No AMI set and use_latest_ami is set to false')
```
citruspi:
```
if self.ami is not None or
```

I think this log message is incorrect.

tyr/servers/server.py
```diff
@@ -43,7 +47,7 @@ def __init__(self, group=None, server_type=None, instance_type=None,
                  block_devices=None, chef_path=None, subnet_id=None,
                  dns_zones=None, ingress_groups_to_add=None,
                  ports_to_authorize=None, classic_link=False,
-                 add_route53_dns=True):
+                 add_route53_dns=True, platform=None, use_latest_ami=False):
```
citruspi: You'll need to add these arguments to all the classes which inherit from `Server`, otherwise they won't be exposed. See [this][0] and [this][1].

[0]: https://github.com/hudl/Tyr/issues/117
[1]: https://github.com/hudl/Tyr/pull/115#issuecomment-232234407

citruspi:
This looks fine, aside from the unaddressed comments that I've made. The pull request also contains several of the issues that I've noted in #111 (and which have been fixed in that pull request).

Similarly, that pull request contains several issues which I've noted and which have been fixed in this pull request.

If they're going to be merged together, I'd recommend merging one branch into another to consolidate the fixes and closing the other pull request.

tyr/servers/server.py
```diff
@@ -65,7 +69,26 @@ def __init__(self, group=None, server_type=None, instance_type=None,
         self.classic_link = classic_link
         self.add_route53_dns = add_route53_dns
         self.ebs_optimized = False
+        self.platform = platform
         self.create_alerts = False
+        self.use_latest_ami = use_latest_ami
+
+    def get_latest_ami(self, ami=None, platform="linux"):
+        if self.ami is not None or self.use_latest_ami is False:
```
citruspi: I think you might be confusing `self.ami` and `ami`

tyr/servers/server.py
```diff
@@ -65,7 +69,26 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
+            self.log.info('AMI has been set and use_latest_ami is set to false')
```
citruspi: This log message is still wrong.

tyr/servers/server.py
citruspi: _bump_

tyr/servers/nginx/server.py
```diff
@@ -32,8 +38,28 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
+
+            validation_key_path = os.path.expanduser('~/.chef/chef-validator.pem')
+            validation_key_file = open(validation_key_path, 'r')
+            validation_key = validation_key_file.read()
+
+            return user_data.format(validation_key=validation_key)
+
+        except IOError:
+            # Handle error reading file
+            pass
```
citruspi: This exception needs to be better handled; the user should be provided with some kind of information about the problem.

tyr/servers/nginx/server.py
```diff
@@ -32,8 +38,28 @@ def __init__(self, group=None, server_type=None, instance_type=None,
...
+
+    @property
+    def user_data(self):
+        # read in userdata file
+        user_data = None
+        try:
+            f = data_file('user_data_chef_provision')
+            user_data = f.read()
+
+            validation_key_path = os.path.expanduser('~/.chef/chef-validator.pem')
```
citruspi: This should use `os.path.join` and `self.chef_path`.

tyr/lifecycle/asgevent.py
```diff
@@ -0,0 +1,38 @@
...
+import logging
+
+class ASGEvent(object):
+    def __init__(self, hook_name=None,
+                asg_name=None,
+                transition='autoscaling:EC2_INSTANCE_LAUNCHING',
+                role_name=None):
+
+        logging.getLogger('botocore').setLevel(logging.CRITICAL)
+        self.log = logging.getLogger('Tyr.Lifecycle.ASGEvent')
```
citruspi: Take a look at [this][0].

[0]: https://github.com/hudl/Tyr/blob/master/tyr/servers/server.py#L77

scripts/build-nginx-service
```diff
@@ -0,0 +1,58 @@
...
+        desired=0,
+        max=9,
+        min=0,
+        health_check_grace_period=900,
+        default_cooldown=900):
+
+    region = 'us-east-1'
+    #prod_subnet_ids=['subnet-e7156e90', 'subnet-f42eb1ad', 'subnet-bfe16094']
+
+    tags = list()
```
citruspi:
```
tags = [
    {'foo': 'bar'}
]
```

citruspi:
LGTM.

PR #115, opened 2016-07-01
Beta support chef12

Z3r0Sum:
# Start adding compatibility/flexibility for different Chef Servers

## Part 1

- Introduce a new attribute Tyr `chef_server_url`
- If not passed in, the attribute will be set to the correct proxy
  server url that allows access to the chef12 server that resides in
  the internal VPC.
- Also need to account for the variance in the
  'validation_client_name' within the `client.rb` file.  Make sure
  that if we're using the legacy chef11 server that it gets set
  correctly.
- Add the environment in the `client.rb` file.  For now using the
  `environment` attribute in the Server Class.
- Start using the `chef_path` attribute
- Fix issue #117 

### Testing

- [x] Stage Server built with Tyr talks to new chef12 server
- [x] Internal VPC Server built with Tyr talks to new chef12 server
- [x] Prod VPC Server built with Tyr talks to new chef12 server
- [x] Old Chef server tested

## Part 2

Implement a new way to do chef attributes without the use of `pyChef`.  The `pyChef` library was fairly inadequate with how it dealt with ACLs during a node's initial run (bootstrapping).  As such, created a new implementation around setting chef node attributes.  We have a top level dictionary value: `CHEF_ATTRIBUTES` in the `Server` class.  Additionally, each class can take advantage of a `set_chef_attributes` function to set specific attributes related to the Server Type as well as include the base `Server` class attributes (as of now, nothing of value is done at that level).  This function would then just be applied during configuration time for the node.

Here's an example from the `MongoDataNode` class:

```
    def set_chef_attributes(self):
        super(MongoDataNode, self).set_chef_attributes()
        ebs_volumes = [
            {
                'user': 'mongod',
                'group': 'mongod',
                'size': self.data_volume_size,
                'iops': self.data_volume_iops,
                'device': '/dev/xvdf',
                'mount': '/volr'
            },
            {
                'user': 'mongod',
                'group': 'mongod',
                'size': self.journal_volume_size,
                'iops': self.journal_volume_iops,
                'device': '/dev/xvdg',
                'mount': '/volr/journal'
            },
            {
                'user': 'mongod',
                'group': 'mongod',
                'size': self.log_volume_size,
                'iops': self.log_volume_iops,
                'device': '/dev/xvdh',
                'mount': '/mongologs',
            }
        ]

        if self.ephemeral_storage == []:
            ebs_volumes.append({
                'user': 'root',
                'group': 'root',
                'size': 8,
                'iops': 24,
                'device': '/dev/xvdc',
                'mount': '/media/ephemeral0'
            })

            self.log.debug('No instance storage; including swap device')

        self.CHEF_ATTRIBUTES['hudl_ebs'] = {'volumes': ebs_volumes}
        self.log.info('Configured the hudl_ebs.volumes attribute')

    def configure(self):

        super(MongoDataNode, self).configure()
        self.set_chef_attributes()
        . . .
```

#### Update and build a node from each class
- [x] Mongo Data Node
- [x] Mongo Config Server
- [x] Mongo Arbiter
- [x] Rabbit
- [x] Nginx --- not creating attributes in the class
- [x] Postfix
- [x] Kafka
- [x] Zookeeper
- [x] Solr 
- [x] Elasticsearch
- [x] Cache

~~- [ ] EurekaServer --- doesn't seem like it's being used?~~

# Issue uncovered

- Probably need another proxy until internal/stage VPC are peered - otherwise there will be connectivity issues when using the stage VPC and trying to talk to the new chef12 server.


tyr/servers/server.py
```diff
@@ -439,21 +455,38 @@ def user_data(self):
...
 curl -L https://www.opscode.com/chef/install.sh | bash;
 yum install -y gcc
 chef-client -S 'http://chef.app.hudl.com/' -N {name} -L {logfile}
 --===============0035287898381899620==--
 """
 
-        validation_key_path = os.path.expanduser('~/.chef/chef-validator.pem')
-        validation_key_file = open(validation_key_path, 'r')
+        try:
+            validation_key_path = os.path.expanduser(self.chef_path +
```
citruspi: This should probably use `os.path.join`.

citruspi:
This looks good to me, the only change I'd suggest would be the order of the arguments in each of the server classes. I've tried to maintain the same order so that when developers use tools like `bpython`, the order of the arguments that pop up is similar each time.

Right now, every class the inherits from the `Server` class lists the same arguments in the same order, from `group` to `dns_zones` and then lists the more specific ones.

For example, from `master`:

```python
    # cache

    def __init__(self, group=None, server_type=None, instance_type=None,
                 environment=None, ami=None, region=None, role=None,
                 keypair=None, availability_zone=None, security_groups=None,
                 block_devices=None, chef_path=None, subnet_id=None,
                 dns_zones=None, couchbase_version=None,
                 couchbase_username=None, couchbase_password=None,
                 bucket_name=None):
```

```python
    # kafka

    def __init__(self, group=None, server_type=None, instance_type=None,
                 environment=None, ami=None, region=None, role=None,
                 keypair=None, availability_zone=None, security_groups=None,
                 block_devices=None, chef_path=None, subnet_id=None,
                 dns_zones=None, volume_count=None, volume_size=None,
                 volume_type=None, logdir_root=None, zookeeper_connection=None):
```

```python
    # postfix

    def __init__(self, group=None, server_type=None, instance_type=None,
                 environment=None, ami=None, region=None, role=None,
                 keypair=None, availability_zone=None,
                 security_groups=None, block_devices=None,
                 chef_path=None, subnet_id=None, dns_zones=None,
                 mail_name=None):
```

citruspi:
Looks good to me, thanks!

Z3r0Sum:
@citruspi I was able to get this working today - I stubbed out some test files in the global dictionary at the Server class level.  Chef ran successfully for both Chef 11 and Chef 12 - was wondering if you could take a quick look at the latest implementation before I start changing the rest of the classes within Tyr.

citruspi:
I'm not sure if you got around to looking into it after we talked, but if it's possible to use something like `printf` instead of using `python -c` in the user data, I would prefer that. `echo` could also technically work since JSON doesn't require line breaks.

Z3r0Sum:
@citruspi I tried a bunch of different ways with `echo`, but always wound up with a empty file or incorrectly formatted in terms of quotes.  I can look into `printf`.

Z3r0Sum:
@citruspi do you anticipate issues when I spin up a solr server in test to validate these changes?

citruspi:
Code Review checklist for my sanity:

- [ ] setup.py
- [ ] tyr/servers/cache/server.py
- [ ] tyr/servers/elasticsearch/server.py
- [ ] tyr/servers/eureka/server.py
- [ ] tyr/servers/kafka/broker.py
- [ ] tyr/servers/mongo/arbiter.py
- [ ] tyr/servers/mongo/config.py
- [ ] tyr/servers/mongo/data.py
- [ ] tyr/servers/mongo/data_warehousing.py
- [ ] tyr/servers/mongo/member.py
- [ ] tyr/servers/mongo/node.py
- [ ] tyr/servers/nginx/server.py
- [ ] tyr/servers/postfix/master.py
- [ ] tyr/servers/rabbit/server.py
- [ ] tyr/servers/scripts/server.py
- [ ] tyr/servers/server.py
- [ ] tyr/servers/solr/master.py
- [ ] tyr/servers/solr/slave.py
- [ ] tyr/servers/zookeeper/server.py

PR #120, opened 2016-08-04
Amazon Linux 2016.03 Support

citruspi:
This updates the default AMI to `Amazon Linux 2016.03` (previously `Amazon Linux 2015.09`) and updates the default Linux cloud config user data to use the correct repository version (previously `2015.03`).


Z3r0Sum:
LGTM

PR #123, opened 2016-08-09
Gamma-230: added appNames to user data

wliao008:
Gamma is going to create different ASG for hudl, admin and queuedservices, when these server spins up it should install only the respective app to those servers, so a QS shouldn't be installing admin, and admin shouldn't be installing hudl, etc. The AppNames will dictate what apps can be installed on the server.

**Rollout**
Add `TreatAsNonMultiverseMonolith` tag to existing monolith ASG with value `yes` iif launch configuration for the ASG is changed.

### DIFF
--- a/data/user_data_base.ps2
+++ b/data/user_data_base.ps2
@@ -199,10 +199,10 @@ function Get-MyName {
     return $nameTag.Value
 }
 
-function Get-AppNames {
+function Get-Apps {
     $instanceId = (New-Object System.Net.WebClient).DownloadString("http://169.254.169.254/latest/meta-data/instance-id")
-    $appNameTag =  Get-EC2Tag | ` Where-Object {$_.ResourceId -eq $instanceId -and $_.Key -eq 'AppNames'}
-    return $appNameTag.Value
+    $appsTag =  Get-EC2Tag | Where-Object {$_.ResourceId -eq $instanceId -and $_.Key -eq 'Apps'}
+    return $apps.Value
 }
 
 $TextInfo = (Get-Culture).TextInfo
@@ -228,7 +228,7 @@ $File = Get-Content "c:\hudl\config\servername.txt"
 New-EC2Tag -Region $region -ResourceId $instanceId -Tag @{key="Name"; value="$servername"}
 
 $roleAttributes = Get-HudlRoleAttributes
-$appNames = Get-AppNames
+$apps = Get-Apps
 
 # Set role tag if not set.
 $chef_tag_role = Get-EC2Tag -Filters @( @{Name="key"; values="Role"}, @{Name="resource-id"; Values=$instanceId})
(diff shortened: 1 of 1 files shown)


data/user_data_base.ps2
```diff
@@ -199,6 +199,12 @@ function Get-MyName {
     return $nameTag.Value
 }
 
+function Get-AppNames {
+    $instanceId = (New-Object System.Net.WebClient).DownloadString("http://169.254.169.254/latest/meta-data/instance-id")
+    $appNameTag =  Get-EC2Tag | ` Where-Object {$_.ResourceId -eq $instanceId -and $_.Key -eq 'AppNames'}
```
chrisgilbert: Does the ` belong there?  I guess you probably took out the line break and it doesn't actually matter, but thought I'd mention it.

chrisgilbert:
Looks good 🎯

data/user_data_base.ps2
```diff
@@ -199,6 +209,12 @@ function Get-MyName {
     return $nameTag.Value
 }
 
+function Get-Apps {
+    $instanceId = (New-Object System.Net.WebClient).DownloadString("http://169.254.169.254/latest/meta-data/instance-id")
+    $appsTag =  Get-EC2Tag | Where-Object {$_.ResourceId -eq $instanceId -and $_.Key -eq 'Apps'}
```
jamiegs: Does this mean we're going to have to have a new apps tag added for all ASGs? if the tag doesn't exist can we default to using hudl-{group}?

wliao008: The `Apps` tag dictates the value in `AppNames` in outpost2.config, as long as `AppNames` is null or empty, it'll be ignored by outpost2 and default to using hudl-{group}. But this gets me thinking there might be a subtle situation here with this approach of whitelisting the apps, mainly once this is non-empty, outpost2 will start filtering out apps according to what's specified, so for a hudl-queuedservices server for eg, in outpost2.config: `"AppNames": = ["hudl-queuedservices"]`, this will fail when we try to deploy outpost2 to this server as outpost2 isn't listed :( 

So if `Apps` is used, it really need to spell out all the apps that are going to be deployed to it, in this case, the `Apps` tag will be `alyx-outpost; hudl-queuedservices`, that creates `"AppNames": = ["alyx-outpost", "hudl-queuedservices"]`. I'm gonna test this out.

jamiegs: Could outpost just always allow `alyx-outpost` rather than having to specifically add it in to the apps list?

wliao008: We could special case `alyx-outpost` in outpost2, that can be done in outpost2 pr, until that is done though, we have to specifically add `alyx-outpost` to the list.

wliao008: Added https://github.com/hudl/Outpost2/pull/77 to make `alyx-outpost` optional

jamiegs: Thanks for doing that.

PR #124, opened 2016-08-11
Added IAM inline policy 

acsportstec:
Added IAM inline policy allow-update-route53-env for dvms to nginx clusters


malkolm:
Looks good 👍

PR #125, opened 2016-08-17
wait for tags on server initialization, save chef command to reusable script file.

jamiegs:
We had an issue were servers were not waiting for tags to be created and were getting the wrong version of mongos installed. 

This will have them wait upto 5 minutes for a tag.

This also stores the chef run command to a script and runs it from there so it could be easily re-ran later.

### DIFF
--- a/data/user_data_base.ps2
+++ b/data/user_data_base.ps2
@@ -213,7 +213,7 @@ function Get-MyName {
 $timeout = new-timespan -Minutes 5
 $sw = [diagnostics.stopwatch]::StartNew()
 
-### Wait until tags are set or it's longer than 5 minutes
+### Wait until tags are set and it's been waiting for less than 5 minutes
 Do {
     Write-Output "Waiting for tags to be added"
 } while ((-not (Check-AnyTagExists) ) -and ($sw.elapsed -lt $timeout))
--- a/data/user_data_base.ps2
+++ b/data/user_data_base.ps2
@@ -8,6 +8,7 @@ Add-Type -Path "C:\Program Files (x86)\AWS Tools\PowerShell\AWSPowerShell\AWSSDK
 
 $AwsTools = Get-Module "AWSPowerShell"
 
+$RUN_CHEF_SCRIPT = "c:\\hudl\\run-chef.ps1"
 $BASE_DIR="C:\hudl"
 $CONFIG="$BASE_DIR\config"
 $LOG_DIR="$BASE_DIR\script-logs"
@@ -146,6 +147,16 @@ function New-ServerName {
     return $myName
 }
 
+function Check-AnyTagExists {
+    $tags = Get-EC2Tag -Filters @(@{Name="resource-id"; Values=$instanceId})
+    if ($tags) {
+        return $true
+    } else {
+        return $false
+    }
+}
+
+
 function GetNewId
 {
     $ticks = [System.DateTime]::Now.Ticks;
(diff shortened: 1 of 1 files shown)


data/user_data_base.ps2
```diff
@@ -199,13 +210,28 @@ function Get-MyName {
     return $nameTag.Value
 }
 
+$timeout = new-timespan -Minutes 5
+$sw = [diagnostics.stopwatch]::StartNew()
+
+### Wait until tags are set or it's longer than 5 minutes
+Do {
+    Write-Output "Waiting for tags to be added"
+} while ((-not (Check-AnyTagExists) ) -and ($sw.elapsed -lt $timeout))
```
citruspi: According to the comment on line 216, this should be an `-or`, not an `-and`.

data/user_data_base.ps2
```diff
@@ -146,6 +147,16 @@ function New-ServerName {
     return $myName
 }
 
+function Check-AnyTagExists {
+    $tags = Get-EC2Tag -Filters @(@{Name="resource-id"; Values=$instanceId})
```
Z3r0Sum: Were you going to use the Group tag here?

data/user_data_base.ps2
```diff
@@ -199,13 +210,28 @@ function Get-MyName {
...
 
+$timeout = new-timespan -Minutes 5
+$sw = [diagnostics.stopwatch]::StartNew()
+
+### Wait until tags are set or it's longer than 5 minutes
+Do {
+    Write-Output "Waiting for tags to be added"
+} while ((-not (Check-AnyTagExists) ) -and ($sw.elapsed -lt $timeout))
+
+if ($sw.elapsed -ge $timeout) {
```
citruspi: Instead of checking if the elapsed time is greater than the timeout, would it make sense to just check the value of `Check-AnyTagExists`? Although unlikely, it'd be possible for us to exit the loop above at something like 4 minutes and 59 seconds (assuming the `-and` becomes an `-or` and the tags get set at that point) and then check if the elapsed time is greater than the timeout immediately after the 5 minute mark. 

So we'd have exited the loop because the tags got set, but the log message would say that it timeout while waiting for the tags to get set.

data/user_data_base.ps2
jamiegs: group doesn't get created until this script runs.. so no. I was planning on using the aws:autoscaling:groupName, but I figured just wait until any tag is created.

data/user_data_base.ps2
```diff
@@ -389,9 +415,15 @@ $attributesContent = $attributesContent + @"
     Write-Output "Beginning write to file"
     [System.IO.File]::WriteAllLines($attributeFile, $attributesContent)
     Write-Output "Chef attributes saved"
-    $env:Path += ";C:\opscode\chef\bin"
     Write-Output "Starting Chef run."
-    Start-Process -FilePath "chef-client" -ArgumentList "-c `"c:\chef\client.rb`" -r `"role[$chef_role]`" -L `"c:\chef\chef-client-initial-run.log`" -j `"c:\chef\attributes.json`"" -Wait -NoNewWindow
+
+    # Create a script for starting chef on disk then run it.
```
wliao008: Just curious why write it to disk first vs Start-Process?

data/user_data_base.ps2
jamiegs: it's a `while less than 5 minutes`..  and is right,  I should change the comment I guess.

citruspi: Ah, gotcha. In that case, I retract my other comment. :)

data/user_data_base.ps2
jamiegs: so if chef fails we can just rerun c:\hudl\run-chef.ps1 rather than having to know the command to run it.

data/user_data_base.ps2
Z3r0Sum: 👍

data/user_data_base.ps2
```diff
@@ -199,13 +210,28 @@ function Get-MyName {
...
 
+$timeout = new-timespan -Minutes 5
+$sw = [diagnostics.stopwatch]::StartNew()
+
+### Wait until tags are set and it's been waiting for less than 5 minutes
+Do {
+    Write-Output "Waiting for tags to be added"
+} while ((-not (Check-AnyTagExists) ) -and ($sw.elapsed -lt $timeout))
+
+if (Check-AnyTagExists) {
```
citruspi: I think you're missing a `-not`.

jamiegs: :facepalm: good catch

jamiegs: I just swapped the log messages around instead.

PR #126, opened 2016-08-22
Fetch Windows User Data From GitHub

citruspi:
There's a limit of 16 KB for user data in EC2, a limit which we are now hitting with the `user_data_base.ps2` script for Windows instances.

I've added a small "helper" user data script which just fetches the full script from GitHub and executes it. I've also updated the IIS node class to use this new script, so any new Auto-Scaling Groups will automatically use this script.

### DIFF
--- a/data/user_data_fetch_github.ps2
+++ b/data/user_data_fetch_github.ps2
@@ -1,0 +1,12 @@
+<powershell>
+# Fetch Windows User Data from GitHub
+
+$USER_DATA_URI = "https://github.com/hudl/Tyr/raw/master/data/user_data_base.ps2"
+$USER_DATA_FILE_PATH = "c:\\hudl\\user_data_base.ps1"
+
+Write-Output "Fetching user data from GitHub"
+Invoke-WebRequest $USER_DATA_URI -OutFile $USER_DATA_FILE_PATH
+
+Write-Output "Executing user data"
+Invoke-Expression $USER_DATA_FILE_PATH
+</powershell>
--- a/data/user_data_fetch_github.ps2
+++ b/data/user_data_fetch_github.ps2
@@ -1,8 +1,13 @@
 <powershell>
 # Fetch Windows User Data from GitHub
 
+$HUDL_DIRECTORY = "c:\\hudl"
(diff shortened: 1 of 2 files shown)


data/user_data_fetch_github.ps2
```diff
@@ -0,0 +1,12 @@
+<powershell>
+# Fetch Windows User Data from GitHub
+
+$USER_DATA_URI = "https://github.com/hudl/Tyr/raw/master/data/user_data_base.ps2"
+$USER_DATA_FILE_PATH = "c:\\hudl\\user_data_base.ps1"
```
jamiegs: I would not assume that c:\hudl exists.

data/user_data_fetch_github.ps2
```diff
@@ -0,0 +1,17 @@
...
+$HUDL_DIRECTORY = "c:\\hudl"
+$USER_DATA_URI = "https://s3.amazonaws.com/hudl-chef-artifacts/windows/user_data_base.ps2"
+$USER_DATA_FILE_PATH = "c:\\hudl\\user_data_base.ps1"
+
+if(!(Test-Path -Path $HUDL_DIRECTORY )){
+    New-Item -ItemType directory -Path $HUDL_DIRECTORY
+}
+
+Write-Output "Fetching user data from GitHub"
+Invoke-WebRequest $USER_DATA_URI -OutFile $USER_DATA_FILE_PATH
```
jamiegs: You might need to add -UseBasicParsing to this.  I'd guess that 'non-basic' parsing probably isn't possible from the user account that runs this script.

data/user_data_fetch_github.ps2
```diff
@@ -0,0 +1,17 @@
...
+
+Write-Output "Executing user data"
+Invoke-Expression $USER_DATA_FILE_PATH
```
jamiegs: Could you check if t he results of running the user_data_base.ps1 get logged to the same log file? I'm guessing not, so we should log it somewhere.

citruspi: I checked, it gets logged to the same file.

PR #127, opened 2016-09-12
Fix dir creation to continue if they already exist

chrisgilbert:
Since switching to this file being downloaded via S3, one of the directories already exists.  This causes an error to be thrown on creation which was silently caught, stopping the other dir creations from taking place.

### DIFF
--- a/data/user_data_base.ps2
+++ b/data/user_data_base.ps2
@@ -31,7 +31,7 @@ $REGIONS = @{'ap-northeast-1' = 'apne1';
 
 Set-DefaultAWSRegion -Region $region
 
-function Create-DirIfNotPresent {
+function New-DirIfNotPresent {
 Param (
     [Parameter(Mandatory=$True )] [string]$DirPath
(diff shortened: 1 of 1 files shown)


data/user_data_base.ps2
```diff
@@ -30,11 +30,19 @@ $REGIONS = @{'ap-northeast-1' = 'apne1';
              'us-west-2'= 'usw2'}
 
 Set-DefaultAWSRegion -Region $region
-try {
-  New-Item $BASE_DIR -type directory
-  New-Item $CONFIG -type directory
-  New-Item $LOG_DIR -type directory
-} catch {}
+
+function Create-DirIfNotPresent {
```
jamiegs: Create is not a approved powershell verb, might want to use `new`

chrisgilbert: OK, thanks 👍

citruspi:
Looks good to me. 👍

PR #128, opened 2016-09-15
use chef-client 12.13.47

jamiegs:
Chef-client 12.14.60 breaks some places we're using yum.  This is a temporary change to using the older version of chef-client which works until we can get our cookbooks updated.


devlife:
:shipit:

citruspi:
Looks good to me. 👍 

I'd just suggest filing an issue to revert these changes so that we don't forget.

PR #130, opened 2016-09-20
remove monolith as multiverse code

jamiegs:
Switch to using a `chef:hudl:treatMonolithAsMultiverse` tag for treatMonolithAsMultiverse rather than using special user-data code.

This required some changes to role_windows_webserver https://github.com/hudl/role_windows_webserver/pull/49 but it makes admin and monolith webapp servers act more like typical multiverse servers and reduced the user data code significantly.

### DIFF
--- a/data/user_data_base.ps2
+++ b/data/user_data_base.ps2
@@ -87,7 +87,6 @@ function Get-HudlRoleAttributes {
     $role = Get-MyRole
 
     #reset values
-    $treatMonolithAsMultiverse = $false
     $environment = $environmentPrefix = $serverRole = $group = ""
 
     $roleSplit = $role -Split '-'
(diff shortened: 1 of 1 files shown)


data/user_data_base.ps2
```diff
@@ -104,36 +103,18 @@ function Get-HudlRoleAttributes {
         $environmentPrefix = $roleSplit[0]
         $serverRole = $roleSplit[1]
         $group = "monolith"
-        # special case for queuedservices and admin here
-        if ($serverRole -eq "queues" -or $serverRole -eq "admin") {
```
wliao008: Is this being handled? Queue servers are using `RoleWeb` now.

jamiegs: I haven't done it for queued services yet, but it'd be done in the cookbook instead like how admin is.

PR #132, opened 2016-12-13
pass cloud manager values through to chef

jamiegs:
Passes values needed to setting up cloud manager through to chef.


jamiegs:
We don't use automation agent any longer, so these changes are not necessary

PR #137, opened 2017-03-09
Set Chef Attributes for Zuun

citruspi:
- Removes the unsupported `mongodb_version` argument
- Removes the unused `MongoDataWarehousingNode` class
- Sets the Chef attributes necessary for integration with Zuun


chrisgilbert:
Hey @citruspi , there's a couple of bugs in the master branch that @atharvai hit earlier today (like the volume size not defaulting correctly).  I fixed that in here.  Any problems with merging this to master now?

chrisgilbert:
This now should run the zuun:configure recipe and creates/updates the databag item zuun requires to install mongo.

PR #138, opened 2017-04-17
switch linux userdata to chef12

jamiegs:
This switches the linux userdata (used in Craft and Nginx ASGs) to use chef12 rather than chef11

* Switches the chef server path
* Switches to use uses lowercase roles, our chef12 server is using lowercase role names.
* Updates chef-client from 12.13.37 to 12.19.36 
* Disables ssl_verify_mode, since our chef12 server doesn't have SSL setup.
* switches the validator name from chef-validator to hudl-validator
* switches to pull the chef validator from s3 rather than being read into a template.
* adds support for reading chef attributes from ec2 tags. the tag must start with `chef:hudl` to support changes in https://github.com/hudl/role_nginx/pull/45

### DIFF
--- a/data/user_data_chef_provision
+++ b/data/user_data_chef_provision
@@ -15,6 +15,8 @@ Content-Transfer-Encoding: 7bit
 Content-Disposition: attachment; filename="user-script.txt"
 #!/bin/bash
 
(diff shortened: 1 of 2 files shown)


chrisgilbert:
Looks good to me Jamie 💃 👍

acsportstec:
:+1: - Yep, these changes shouldn't affect us

PR #148, opened 2017-09-03
Resolve Dependencies Once

citruspi:
A common pattern in Tyr is inheriting from a parent Server type and then in the `configure` method:
- Calling `super(Type).configure()`
- Modifying the server's configuration (e.g. IAM policies, security groups, etc)
- Re-resolving these dependencies

However, this means that when provisioning e.g. a `MongoDataNode` you end up setting the Chef attributes 4 times, resolving the security groups twice, and resolving the IAM policies thrice. That's a lot of wasted time.

This moves
- resolution of IAM stuff
- resolution of security groups
- resolution of ingress groups to add
- setting of Chef attributes

to a method called `resolve_dependencies` which should be invoked _once_ before invoking `launch` and after every invocation of `configure` has been completed. 

I've made the necessary adjustments to `autorun` in `Server` and `provision` in the `IIS` and `Nginx` clusters to invoke it.

Essentially this significantly reduces the time spent between `establish_logger` and `launch` and removes a number of API requests for information that we already have.


citruspi:
In addition, the `resolve_dependencies` method formats each IAM policy, security group, and ingress group to add. So instead of writing
```
self.ingress_groups_to_add.extend([		
    "{env}-web".format(env=env_prefix),		
    "{env}-nginx".format(env=env_prefix),		
    "{env}-queueproc-jobs".format(env=env_prefix)		
])
```
you can just write
```
self.ingress_groups_to_add.extend([		
    '{env}-web',
    '{env}-nginx',
    '{env}-queueproc-jobs'
])
```
The following variables are available during formatting:

| Variable | Value | Example |
|:---------|:------|:---------|
| `environment` | `environment` | `prod` |
| `env` | `environment[0]` | `p` |
| `group` | `group` | `speedtest` |
| `type` | `server_type` | `mongo` |
| `envcl` | `{env}-{group}-{type}` | `p-speedtest-mongo` |

I've set the default security groups for each server to
```
['management', '{env}-{type}-management',  '{envcl}', 'chef-nodes']
```

PR #154, opened 2021-09-03
Allow for specifying the chef client version in the Tyr request

lewislabs:
## Core change

AWS released a new version of a Ruby gem (`aws-eventstream`) which only supports Ruby >=v2.3. Chef-client uses this gem for things when provisioning a new instance. So the gem upgrade effectively borked chef-client for newly provisioned instances (via Tyr) 

v12.14 of Chef included an upgrade of Ruby and it's only a minor bump away from the version we've been using. 

This PR makes this version configurable, but defaults to the 12.14 version. 

It's been tested so far with provisioning a MongoDb router node.

## Other improvements
- Add Changelog
- Add README with pointer to docs in Sync
- Snapshot code from Hudl's `run_instance_handler` Lambda

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,8 @@
+# CHANGELOG
+
+<https://keepachangelog.com/en/1.0.0/>
+
+## [2.2.0]
+
+- Allow for specifying the Chef client version in the Tyr request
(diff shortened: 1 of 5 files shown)


dnordgren (approved):
TIL Tyr is open source.

Seems legit to me!

andypryor (approved):
lgtm

chrisgilbert (approved):

